PR #15, opened 2021-11-30
certs are now passed in already read, not file location

eoinmcafee00:


cli/certs/certs.go
```diff
@@ -166,3 +174,19 @@ func newCertificate(org string) (*x509.Certificate, error) {
 		BasicConstraintsValid: true,
 	}, nil
 }
+
+func ReadCerts(caCertFileLocation, certFileLocation, certKeyLocation string) (*certList, error) {
+	caCertFile, err := os.ReadFile(caCertFileLocation)
+	if err != nil {
+		return nil, err
+	}
+	certFile, err := os.ReadFile(fmt.Sprintf(certFileLocation))
```
tphoney, line 183: do we need `fmt.Sprintf(certFileLocation)`

tphoney (changes requested):

cli/certs/certs.go
eoinmcafee00: done , good catch

tphoney (approved):

PR #19, opened 2021-12-02
update setup response

eoinmcafee00:


api/api.go
```diff
@@ -34,7 +34,10 @@ type (
 		IsDir bool   `json:"is_dir,omitempty"`
 	}
 
-	SetupResponse struct{}
+	SetupResponse struct {
```
shubham149, line 37: This is incorrect. Lite-engine does not return instance_id and ip_address as response. That is returned by runner api. We need to revert this PR.

PR #35, opened 2021-12-20
Add support for output vars for windows and proxy env for ti & log svc call

shubham149:


pipeline/runtime/common.go
```diff
@@ -24,15 +24,26 @@ func getNudges() []logstream.Nudge {
 	}
 }
 
-func getOutputVarCmd(outputVars []string, outputFile string) string {
+func getOutputVarCmd(entrypoint, outputVars []string, outputFile string) string {
 	cmd := ""
 	for _, o := range outputVars {
-		cmd += fmt.Sprintf(";echo \"%s $%s\" >> %s", o, o, outputFile)
+		env := fmt.Sprintf("$%s", o)
+		if isPowershell(entrypoint) {
```
vistaarjuneja, line 31: What if we don't have any entrypoint (running directly on VM)?

shubham149: This method won't be called. Output variable does not make sense if entrypoint is not present. It is a plugin step instead.

shubham149: it errs out here: https://github.com/harness/lite-engine/pull/35/files#diff-b07c32047a7ce106899517b16b9e24c271e15b45b78e514c4dd34b5d790e570fL24

vistaarjuneja (approved):

PR #37, opened 2021-12-21
(feat) implement file/folder creation on setup

tphoney:


marko-gacesa (approved):

engine/engine.go
```diff
@@ -40,6 +41,12 @@ func NewEnv(opts docker.Opts) (*Engine, error) {
 }
 
 func (e *Engine) Setup(ctx context.Context, pipelineConfig *spec.PipelineConfig) error {
+	// create global files and folders
```
shubham149, line 44: Create files in handler/setup.go.

engine/engine.go
```diff
@@ -85,6 +92,11 @@ func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*r
 	}
 	step.Envs = envs
 	step.WorkingDir = pathConverter(step.WorkingDir)
+	// create files or folders specific to the step
+	fileFolderCreationErr := fileFolderCreation(step.Files)
```
shubham149, line 96: Move this to handler/step.go

engine/engine.go
```diff
@@ -97,6 +109,44 @@ func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*r
 	return exec.Run(ctx, step, output)
 }
 
+func fileFolderCreation(paths []*spec.File) error {
+	for _, f := range paths {
+		if f.Path != "" {
```
shubham149, line 114:
```
if f.Path == "" {
  continue
}

path := f.Path
...
```

This will reduce one nested block and make code easier to read.

engine/engine.go
```diff
@@ -97,6 +109,44 @@ func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*r
...
+			}
+			if f.IsDir {
+				// create a folder
+				if err := os.MkdirAll(path, fs.FileMode(f.Mode)); err != nil {
+					return errors.Wrap(err,
+						fmt.Sprintf("failed to create directory for host path: %s", path))
+				}
+			} else {
+				// create a file
+				newFile, newErr := os.Create(path)
```
shubham149, line 127: Use os.OpenFile and create a file with correct permissions in a single call https://pkg.go.dev/os#OpenFile

engine/spec/spec.go
```diff
@@ -13,6 +13,7 @@ type (
 		Volumes  []*Volume         `json:"volumes,omitempty"`
 		Network  Network           `json:"network"`
 		Envs     map[string]string `json:"envs,omitempty"`
+		Files    []*File           `json:"files,omitempty"`
```
shubham149, line 16: File should not be a part of engine spec ( only needs to be present in api section).

engine/engine.go
marko-gacesa: @shubham149 I disagree. Handler code should deal with HTTP stuff: Parsing request body and parameters, and not the business logic. Architecturally it's better to place this code here - in the engine, because it's manipulating the VM instance.

engine/engine.go
marko-gacesa: Agreed. I'll fix this because TP's on the vacation.

engine/engine.go
marko-gacesa: If the file is supposed to be created as read-only, then we create it with `OpenFile` and the correct permissions (and, of course, drop the call to `Chmod` below), will write to the file fail?

I'm thinking to modify the code. To create the file, write to it, close it, and then call `Chmod` to set the permissions.

engine/spec/spec.go
marko-gacesa: You are right. But I'd leave here too. Because then we will know what files we created if it's ever needed later.

marko-gacesa (approved):

engine/engine.go
```diff
@@ -97,6 +109,52 @@ func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*r
 	return exec.Run(ctx, step, output)
 }
 
+func fileFolderCreation(paths []*spec.File) error {
```
shubham149, line 112: Nit: Change method name to createFiles. 

Folder is also a file in unix.

engine/engine.go
```diff
@@ -97,6 +109,52 @@ func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*r
...
+			if err := os.MkdirAll(path, fs.FileMode(f.Mode)); err != nil {
+				return errors.Wrap(err,
+					fmt.Sprintf("failed to create directory for host path: %s", path))
+			}
+
+			continue
+		}
+
+		// create a file
+		newFile, newErr := os.Create(path)
```
shubham149, line 134: Nit: 
newErr -> err
newFile -> file

engine/engine.go
```diff
@@ -97,6 +109,52 @@ func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*r
...
+		if newErr != nil {
+			return errors.Wrap(newErr,
+				fmt.Sprintf("failed to create file for host path: %s", path))
+		}
+
+		_, writeErr := newFile.WriteString(f.Data)
```
shubham149, line 140: Nit: 


`if _, err = newFile.WriteString(f.Data); err != nil {
 ...
}`

engine/engine.go
```diff
@@ -97,6 +109,52 @@ func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*r
...
+		if writeErr != nil {
+			_ = newFile.Close()
+			return errors.Wrap(writeErr,
+				fmt.Sprintf("failed to write file for host path: %s", path))
+		}
+
+		_ = newFile.Close()
+
+		permErr := os.Chmod(path, fs.FileMode(f.Mode))
```
shubham149, line 149: Nit: 

if err := os.Chmod(path, fs.FileMode(f.Mode)); err != nil {
...
}

shubham149 (approved):

PR #42, opened 2022-01-14
drone support: step exec and output steaming

marko-gacesa:


cli/client/http.go
```diff
@@ -107,6 +107,43 @@ func (c *HTTPClient) RetryPollStep(ctx context.Context, in *api.PollStepRequest,
 	}
 }
 
+func (c *HTTPClient) GetStepLogOutput(ctx context.Context, in *api.StreamOutputRequest, w io.Writer) error {
```
shubham149, line 110: This is not the correct approach. We are already writing the logs to a file. Check: https://github.com/harness/lite-engine/blob/main/logstream/filestore/filestore.go#L70

In case of log streaming, read this file, tail the logs and close the stream when step is complete. It will also ensure that if stream log call fails in between, retry will work. Also, there won't be any waiting for step to start or anything like that.

shubham149: Use this utility for tailing the logs from the file: https://pkg.go.dev/github.com/nxadm/tail#section-readme

Let's discuss the approach on monday.

shubham149 (changes requested):

pipeline/runtime/step_executor.go
```diff
@@ -38,6 +39,7 @@ type StepExecutor struct {
 	engine     *engine.Engine
 	mu         sync.Mutex
 	stepStatus map[string]StepStatus
+	stepOutput map[string]*StepOutput
```
shubham149, line 42: Can we use stepLogOutput instead of stepOutput? step output in CIE refers to output variables.

pipeline/runtime/step_executor.go
```diff
@@ -110,7 +113,122 @@ func (e *StepExecutor) PollStep(ctx context.Context, r *api.PollStepRequest) (*a
...
+	for ts := time.Now(); ; {
+		e.mu.Lock()
+		stepOutput = e.stepOutput[id]
+		e.mu.Unlock()
+
+		if stepOutput != nil {
+			break
+		}
+
+		const timeoutDelay = 5 * time.Second
```
shubham149, line 135: If a step takes a millisecond to complete, it will take atleast 5 seconds in drone since 5 seconds is the minimum time after which logs are sent back.

pipeline/runtime/step_executor.go
```diff
@@ -110,7 +113,122 @@ func (e *StepExecutor) PollStep(ctx context.Context, r *api.PollStepRequest) (*a
...
+	go func() {
+		select {
+		case <-ctx.Done():
+		case <-stepOutput.Done():
+		}
+		close(chData)
+		stepOutput.Unsubscribe(chData)
+	}()
+
+	newOut = chData
```
shubham149, line 166: From this lines, it looks like all the data will be streamed after the step is complete.

pipeline/runtime/step_executor.go
```diff
@@ -110,7 +113,122 @@ func (e *StepExecutor) PollStep(ctx context.Context, r *api.PollStepRequest) (*a
 	return convertStatus(status), nil
 }
 
+func (e *StepExecutor) StreamOutput(ctx context.Context, r *api.StreamOutputRequest) (oldOut []byte, newOut <-chan []byte, err error) {
```
shubham149, line 116: StreamOutput method does not below to this file since it is not related to step execution but for step log streaming. Can you move this method out of this file?

marko-gacesa: `StreamOutput` method is there and should remain there because it's a method of `StepExecutor` type. The struct `StepExecutor` itself and all other its methods are in the same file. Why would we want to move this one to a different file? Besides, a step's output is closely related to its execution, so that's another reason for it to stay where it is. The actual streaming is in an another file: step_output.go

pipeline/runtime/step_executor.go
marko-gacesa: No. How could you conclude that? Quite the opposite. Streaming starts as soon as there is something to stream.

pipeline/runtime/step_executor.go
marko-gacesa: You misunderstood the code. The 5s delay is to abort streaming (if the step has not started in that period), not to start streaming...
Well, I admit it's a hacky solution, but it's there just to ensure that step is started just prior to the streaming. The two requests are started simultaneously by the runner...

pipeline/runtime/step_executor.go
marko-gacesa: Well, we are capturing what a step writes to standard output. So, technically, it should be called 'output'. Output variables should be called outputVars or something like that.

pipeline/runtime/step_executor.go
shubham149: Coupling of log streaming handler and step execution is not a good idea

pipeline/runtime/step_executor.go
shubham149: Logs are captured on standard error as well. So, by that reasoning it should be stepOutput+Error.

Step output can refer to couple of things e.g. output variables, logs etc. Renaming it to avoid confusion.

tphoney (approved):

PR #46, opened 2022-02-08
Bind container port to host port

shubham149:
Port binding support to enable step running on vm to connect to service dependencies via localhost. Step running on container will be able to connect to service dependency via service container name.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -63,6 +63,7 @@ type (
 		MemLimit     int64                `json:"mem_limit,omitempty"`
 		Network      string               `json:"network,omitempty"`
 		Networks     []string             `json:"networks,omitempty"`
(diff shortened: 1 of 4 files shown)


shubham149:
@tphoney https://github.com/harness/lite-engine/pull/46/files#diff-88da8e942ee982cc0e1dfa75788db5f42072de1282697e8fe6ffc23d785d1279R119 is taking care of portBindings. I tested it locally and step running on host was able to connect to the service container.

Regarding the suffix TCP, by default when we specify port binding as 9000:9000, it uses tcp. That's why I used tcp by default.

shubham149:
Addressed the comments.

tphoney (approved):

PR #48, opened 2022-02-22
Moved docker/image package from internal to engine

marko-gacesa:


tphoney:
this makes sense to me.

tphoney (approved):

PR #53, opened 2022-04-14
append usr/local/bin path for osx docker

eoinmcafee00:
This is required for osx support. "docker" command won't work. You need to pass full path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,5 @@
-lite-engine
-lite-engine.exe
+lite-engine*
 .env
 get-docker.sh
 /release/
-/lite-engine-windows-amd64.exe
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@ lite-engine.exe
 .env
 get-docker.sh
 /release/
+/lite-engine-windows-amd64.exe
```
tphoney, line 6: we can summarise ignoring binaries with something like 'lite-engine*'

tphoney (changes requested):

.gitignore
eoinmcafee00: done

tphoney (approved):

PR #67, opened 2022-06-14
fixes issue where default value of true would never get set on MountDockerVolume

eoinmcafee00:


handler/setup.go
```diff
@@ -34,7 +34,7 @@ func HandleSetup(engine *engine.Engine) http.HandlerFunc {
 		state := pipeline.GetState()
 		state.Set(s.Secrets, s.LogConfig, s.TIConfig)
 
-		if s.MountDockerSocket { // required to support m1 where docker isn't installed.
+		if s.MountDockerSocket != nil { // required to support m1 where docker isn't installed.
```
shubham149, line 37: `if s.MountDockerSocket == nil || s.MountDockerSocket  {`

shubham149 (approved):

PR #73, opened 2022-08-08
[CI-4986] Golang change to enable docker runner initialization step

happyhebaby:


api/api.go
```diff
@@ -28,6 +28,14 @@ type (
 		MountDockerSocket *bool             `json:"mount_docker_socket,omitempty"`
 	}
 
+	SetupRequest struct {
```
shubham149, line 31: Why do we require changes to this api?

config/config.go
```diff
@@ -24,7 +24,7 @@ type Config struct {
 	}
 
 	Client struct {
-		Bind       string `envconfig:"HTTPS_BIND" default:":9079"`
+		Bind       string `envconfig:"HTTPS_BIND" default:":3000"`
```
eoinmcafee00, line 27: Why is this changed? This will break everything.

tphoney (changes requested):
This cannot be merged until we have the unit tests running

PR #75, opened 2022-08-17
Upgrade docker client to 20.10.17

shubham149:


go.sum
```diff
@@ -1,336 +1,35 @@
-bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898/go.mod h1:Xbm+BRKSBEpa4q4hTSxohYNQpsxXPbPry4JJWOB3LB8=
```
vistaarjuneja, line 1: Do we not need the `go.sum` file?

shubham149: File is still present. Most of the contents of the file have been removed.

shubham149:
It gives error `Error response from daemon: client version 1.41 is too new. Maximum supported API version is 1.40` in case of gcp windows

tphoney:
> It gives error `Error response from daemon: client version 1.41 is too new. Maximum supported API version is 1.40` in case of gcp windows

You maybe able to use a later version of the windows image, that may allow a later client version

shubham149:
I will try with a lower docker client and update this PR later

PR #85, opened 2022-09-22
(fix): wait for log copying to complete before returning control

vistaarjuneja:


vistaarjuneja:
[Needs testing, will update once done]

engine/docker/docker.go
```diff
@@ -347,15 +347,15 @@ func (e *Docker) tail(ctx context.Context, id string, output io.Writer) error {
...
 	logs, err := e.client.ContainerLogs(ctx, id, opts)
 	if err != nil {
+		logger.FromContext(ctx).WithError(err).
+			WithField("container", id).
+			Debugln("failed to stream logs")
 		return err
 	}
+	defer logs.Close()
+
+	stdcopy.StdCopy(output, output, logs)
```
shubham149, line 357: log the error in case stdcopy returns an err

vistaarjuneja: Done

shubham149 (approved):

PR #98, opened 2022-11-15
Add support for exported env vars

shubham149:
A step can export environment variables. The env variables exported by a step will be appended to all the subsequent steps. 

How a step can export env vars:
Step will write the env variables in a DRONE_ENV file path (DRONE_ENV is an environment variable which will be set to the location of the file on which env vars needs to be set) in .env file format. 
e.g.  `foo=bar`

This functionality is required for implementing github action step. In github action step e.g. java-setup, it installs the java distribution and exports PATH and JAVA_HOME. So, in order for subsequent step to use the installed java, PATH and JAVA_HOME needs to be set. Export environment variable functionality will auto-set the exported env vars to the subsequent steps.

Lite-engine will read the exported env vars and returns it as part PollStepResponse


tphoney (approved):
Excellent work, seems mighty familiar 😃

PR #105, opened 2022-12-27
(fix): Use one bazel query to find all test rules

rutvijmehta-harness:


ti/instrumentation/java/bazel.go
```diff
@@ -81,6 +81,80 @@ func (b *bazelRunner) AutoDetectTests(ctx context.Context, workspace string, tes
 	return tests, nil
 }
 
+func parseBazelTestRule(r string) (ti.RunnableTest, error) {
+	// r = //module:package.class
+	if r == "" {
+		return ti.RunnableTest{}, fmt.Errorf("empty rule")
+	}
+	n := 2
+	if !strings.Contains(r, ":") || len(strings.Split(r, ":")) < n {
```
hemanthmantri, line 90: Nit: can you do regex instead? Same below

ti/instrumentation/java/bazel.go
```diff
@@ -100,6 +174,8 @@ func (b *bazelRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
 		return "echo \"Skipping test run, received no tests to execute\"", nil //nolint:goconst
 	}
 
+	tests = getBazelTestRules(ctx, b.log, tests)
```
hemanthmantri, line 177: Do we still need to do stuff in lines 184-255? I thought we are already getting all the test rules here.

ti/instrumentation/java/bazel.go
rutvijmehta-harness: Comment resolved in other PR.

jamie-harness (approved):

PR #123, opened 2023-02-27
Add output variable support for container less plugins

raghavharness:


pipeline/runtime/common.go
```diff
@@ -70,6 +74,10 @@ func isPython(entrypoint []string) bool {
 	return false
 }
 
+func isContainerlessPlugin(image string, entrypoint []string) bool {
+	return image == "" && len(entrypoint) > 0 && entrypoint[0] == "plugin"
```
shubham149, line 78: this does not seem right. lite-engine should not be aware of plugin binary

pipeline/runtime/common.go
```diff
@@ -118,6 +118,25 @@ func fetchExportedEnvVars(envFile string, out io.Writer) map[string]string {
 	return env
 }
 
+// Fetches output variables exported by the step.
+func fetchOutputVariablesFromEnvFile(outputFile string, out io.Writer) map[string]string {
```
shubham149, line 122: this method is a copy of above one. Can we de-duplicate it?

pipeline/runtime/run.go
```diff
@@ -34,7 +34,9 @@ func executeRunStep(ctx context.Context, engine *engine.Engine, r *api.StartStep
 		return nil, nil, nil, fmt.Errorf("output variable should not be set for unset entrypoint or command")
 	}
 
-	outputFile := fmt.Sprintf("%s/%s.out", pipeline.SharedVolPath, step.ID)
+	outputFile := fmt.Sprintf("%s/%s.env", pipeline.SharedVolPath, step.ID)
```
shubham149, line 37: %s/%s-output.env

shubham149: So, that confusion is not present with DRONE_ENV one

pipeline/runtime/common.go
raghavharness: Let me make a common one.

pipeline/runtime/run.go
raghavharness: Sure

pipeline/runtime/common.go
```diff
@@ -101,19 +101,21 @@ func fetchOutputVariables(outputFile string, out io.Writer) (map[string]string,
...
+// Fetches variable in env file exported by the step.
+func fetchExportedVarsFromEnvFile(envFile string, out io.Writer) map[string]string {
 	log := logrus.New()
 	log.Out = out
 
+	env := make(map[string]string)
+
 	if _, err := os.Stat(envFile); errors.Is(err, os.ErrNotExist) {
-		return nil
+		return env
```
shubham149, line 112: we can return nil here

shubham149: Remove fetchOutputVariables method

raghavharness: okay

shubham149 (approved):

PR #126, opened 2023-03-14
Outputvar fixes

raghavharness:


pipeline/runtime/runtest.go
```diff
@@ -67,10 +67,10 @@ func executeRunTestStep(ctx context.Context, engine *engine.Engine, r *api.Start
 		err = collectionErr
 	}
 
-	exportEnvs := fetchExportedVarsFromEnvFile(exportEnvFile, out)
+	exportEnvs, _ := fetchExportedVarsFromEnvFile(exportEnvFile, out)
```
shubham149, line 70: ditto

pipeline/runtime/common.go
```diff
@@ -69,20 +69,21 @@ func isPython(entrypoint []string) bool {
 }
 
 // Fetches variable in env file exported by the step.
-func fetchExportedVarsFromEnvFile(envFile string, out io.Writer) map[string]string {
+func fetchExportedVarsFromEnvFile(envFile string, out io.Writer) (map[string]string, error) {
```
shubham149, line 72: Why are we returning error if we are not handling in the caller?

pipeline/runtime/run.go
```diff
@@ -49,10 +49,14 @@ func executeRunStep(ctx context.Context, engine *engine.Engine, r *api.StartStep
 		logrus.WithError(rerr).WithField("step", step.Name).Errorln("failed to upload report")
 	}
 
-	exportEnvs := fetchExportedVarsFromEnvFile(exportEnvFile, out)
+	exportEnvs, _ := fetchExportedVarsFromEnvFile(exportEnvFile, out)
```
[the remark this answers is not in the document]
raghavharness: Should I just log here ?

pipeline/runtime/common.go
```diff
@@ -69,20 +69,21 @@ func isPython(entrypoint []string) bool {
...
 	if _, err := os.Stat(envFile); errors.Is(err, os.ErrNotExist) {
-		return nil
+		return nil, err
 	}
 
 	env, err := godotenv.Read(envFile)
 	if err != nil {
-		log.WithError(err).WithField("envFile", envFile).Warnln("failed to read exported env file")
-		return nil
+		content, _ := os.ReadFile(envFile)
```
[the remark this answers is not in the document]
raghavharness: Hmm but this block is itself inside an error cond. If I override the error in this statement then we won't be able to know the actual error.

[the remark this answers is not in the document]
shubham149: you can use a different variable name for err e.g. ferr

pipeline/runtime/runtest.go
```diff
@@ -67,10 +67,10 @@ func executeRunTestStep(ctx context.Context, engine *engine.Engine, r *api.Start
...
 	if len(r.OutputVars) > 0 {
 		if exited != nil && exited.Exited && exited.ExitCode == 0 {
-			outputs := fetchExportedVarsFromEnvFile(outputFile, out)
+			outputs, err := fetchExportedVarsFromEnvFile(outputFile, out) //nolint:govet
```
shubham149, line 73: Do we have to add this code here
if len(r.OutputVars) > 0 {
			// only return err when output vars are expected
			return exited, outputs, exportEnvs, err
		}

pipeline/runtime/common.go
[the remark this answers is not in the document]
raghavharness: okay

pipeline/runtime/runtest.go
raghavharness: If I don't add it then it will fail for plugins. Since plugin can or cannot export output vars.

raghavharness: So only failing when len(r.OutputVars) > 0 but we got error while parsing outputs.

raghavharness: So if you asked for runtest step, we already have a check `if len(r.OutputVars) > 0` here.

shubham149 (approved):

PR #134, opened 2023-04-05
(feat): add noop client for lite engine operations

vistaarjuneja:
We want to be able to support easy scale testing in AWS runner without having to interact with the lite engine. This PR adds a noop client to mock out the lite engine ops

### DIFF
--- a/cli/client/noop.go
+++ b/cli/client/noop.go
@@ -8,6 +8,8 @@ import (
 	"github.com/harness/lite-engine/api"
 )
 
+var _ Client = (*NoopClient)(nil)
+
 type NoopClient struct {
 	stepResponse *api.PollStepResponse // response to return for the step execution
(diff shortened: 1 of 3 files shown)


shubham149 (approved):

cli/client/noop.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/harness/lite-engine/api"
+)
+
+var _ Client = (*NoopClient)(nil)
+
+type NoopClient struct {
+	stepResponse *api.PollStepResponse // response to return for the step execution
+	stepErr      error                 // if there's an error in polling the step response
+	stepExecTime time.Duration         // how long to wait for the step to return back
```
shubham149, line 16: We can add time for setup and destroy to simulate the real env

PR #139, opened 2023-04-26
Add retries on network create call

shubham149:


engine/docker/docker.go
```diff
@@ -93,17 +95,7 @@ func (e *Docker) Setup(ctx context.Context, pipelineConfig *spec.PipelineConfig)
...
-	driver := "bridge"
-	if pipelineConfig.Platform.OS == "windows" {
-		driver = "nat"
-	}
-	_, err := e.client.NetworkCreate(ctx, pipelineConfig.Network.ID, types.NetworkCreate{
-		Driver:  driver,
-		Options: pipelineConfig.Network.Options,
-		Labels:  pipelineConfig.Network.Labels,
-	})
+	err := e.createNetworkWithRetries(ctx, pipelineConfig)
```
vistaarjuneja, line 98: Do we need this if we already have retries at the top level?

shubham149: we don't have retries at the top level.

PR #142, opened 2023-05-05
added support for artifact data

raghavharness:


pipeline/runtime/run.go
```diff
@@ -41,6 +41,9 @@ func executeRunStep(ctx context.Context, engine *engine.Engine, r *api.StartStep
 		step.Command[0] += getOutputVarCmd(step.Entrypoint, r.OutputVars, outputFile)
 	}
 
+	artifactFile := fmt.Sprintf("%s/%s-artifact", pipeline.SharedVolPath, step.ID)
+	step.Envs["PLUGIN_ARTIFACT_FILE"] = artifactFile
```
shubham149, line 45: I will suggest using DRONE_ARTIFACT_FILE instead of PLUGIN_ARTIFACT_FILE

pipeline/runtime/run.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 func executeRunStep(ctx context.Context, engine *engine.Engine, r *api.StartStepRequest, out io.Writer, tiConfig *tiCfg.Cfg) ( //nolint:gocritic
-	*runtime.State, map[string]string, map[string]string, error) {
+	*runtime.State, map[string]string, map[string]string, []byte, error) {
```
shubham149, line 24: Make this change for runtest as well

pipeline/runtime/run.go
raghavharness: For k8 we use PLUGIN_ARTIFACT_FILE right ? Customers can have their plugins which might be generating artifacts and expects PLUGIN_ARTIFACT_FILE as env variable ?

shubham149 (approved):

PR #143, opened 2023-05-05
fix:  [CI-7586]: Increase character limit when parsing Junit XML

hemanthmantri:
The current limit is 2k and a customer complained about the report fields being truncated. This change bumps this limit to 4k.


anurag-harness (approved):
Please make the same change in the addon as well

PR #150, opened 2023-05-12
[fix]: [CI-8043]: Bazel query not working on VMs

anurag-harness:


rutvijmehta-harness (approved):
LGTM. Make sure to test bazel query for both test rules and autodetect tests.

You'll also need to update the UT.

PR #152, opened 2023-05-14
[CI-7842]: Fix import and review comments, add testGlob support

jamie-harness:


ti/instrumentation/utils.go
```diff
@@ -219,18 +219,18 @@ func downloadFile(ctx context.Context, path, url string, fs filesystem.FileSyste
 	// Create the file
 	out, err := fs.Create(path)
 	if err != nil {
-		return fmt.Errorf("could not create directory: %s", err)
+		return err
```
happyhebaby, line 222: It was difficult from the current logging to determine which of the step in this function failed, so I added this additional logging improvement, I think we should keep them to make error diagnosis easier.

ti/instrumentation/python/helper.go
```diff
@@ -31,10 +28,15 @@ func getPythonTestsFromPattern(tests []ti.RunnableTest, workspace, testpattern s
 		if path == "" {
 			continue
 		}
-		test := ti.RunnableTest{
-			Class: path,
+		for _, glob := range testGlobs {
```
rutvijmehta-harness, line 31: Not a blocker but to make it consistent with other implementations, you can do something like this: https://github.com/harness/lite-engine/pull/97/files#diff-b5fb100ad990fe806225fea86c0c6369c47a2ad49b94728e266daf3437161abdR21

ti/instrumentation/python/helper.go
```diff
@@ -43,11 +45,10 @@ func getPythonTestsFromPattern(tests []ti.RunnableTest, workspace, testpattern s
 // In case of errors, return empty list
 func GetPythonTests(workspace string, testGlobs []string) []ti.RunnableTest {
 	tests := make([]ti.RunnableTest, 0)
-	tests, err := getPythonTestsFromPattern(tests, workspace, "test_*.py")
-	if err != nil {
-		return tests
+	if len(testGlobs) == 0 {
+		testGlobs = append(testGlobs, "test_*.py", "*_test.py")
```
rutvijmehta-harness, line 49: Define constant as `defaultTestGlobs = []string{"test_*.py", "*_test.py"}`

Use it there
```suggestion
		testGlobs = defaultTestGlobs
```

ti/instrumentation/python/helper.go
jamie-harness: I will keep this one for now. Because the parseJavaNode part is copied from addon. And is quite a detour because we create a Node and covert to a test

happyhebaby (approved):

rutvijmehta-harness (approved):

PR #153, opened 2023-05-18
Fix python TI non-instrument command

jamie-harness:


ti/instrumentation/instrumentation.go
```diff
@@ -202,7 +202,11 @@ func GetCmd(ctx context.Context, config *api.RunTestConfig, stepID, workspace st
 	}
 
 	// Create the config file required for instrumentation
-	iniFilePath, err := createConfigFile(runner, config.Packages, config.TestAnnotations, workspace, tmpFilePath, fs, log, useYaml)
+	iniFilePath := ""
+	// Python does not use config file now. Will add in the future
+	if !strings.EqualFold(config.Language, "python") {
```
rutvijmehta-harness, line 207: Pass the language to the method and return `iniFilePath` as empty string. Let's not add conditional checks in the `GetCmd` method.

We should also cleanup the runner selection but that's something which can be taken care of later.

rutvijmehta-harness (approved):

ti/instrumentation/instrumentation.go
jamie-harness: ideally, we should have a language runner and build tool runner to extend the language runner. These language level config differences will be in the language runner. Probably will do it some time..

PR #155, opened 2023-05-26
[feat]: [CI-8188]: 0777 permission must be given to /tmp/engine

anurag-harness:
Currently even though we create the dir with 0777, the default umask of 022 creates the dir with 0755. Hence we need to explicitly chmod to achieve this. Please refer to https://harness.atlassian.net/browse/CI-8107 for more details as to why its needed.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -25,6 +25,7 @@ const (
 	DockerSockVolName  = "_docker"
 	DockerSockUnixPath = "/var/run/docker.sock"
 	DockerSockWinPath  = `\\.\pipe\docker_engine`
+	permissions        = 0777
 )
 
 type Engine struct {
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -52,19 +53,22 @@ func (e *Engine) Setup(ctx context.Context, pipelineConfig *spec.PipelineConfig)
 	}
 	// create volumes
 	for _, vol := range pipelineConfig.Volumes {
-		if vol != nil && vol.HostPath != nil {
-			path := vol.HostPath.Path
-			vol.HostPath.Path = pathConverter(path)
+		if vol == nil || vol.HostPath == nil {
```
anurag-harness, line 56: Had to make this change due to failed linter

https://harness.drone.io/harness/lite-engine/930/1/3

engine/engine.go
```diff
@@ -25,6 +25,7 @@ const (
 	DockerSockVolName  = "_docker"
 	DockerSockUnixPath = "/var/run/docker.sock"
 	DockerSockWinPath  = `\\.\pipe\docker_engine`
+	permissions        = 0777
```
hemanthmantri, line 28: s/permissions/Permissions

engine/engine.go
```diff
@@ -52,19 +53,22 @@ func (e *Engine) Setup(ctx context.Context, pipelineConfig *spec.PipelineConfig)
...
+			continue
+		}
+		path := vol.HostPath.Path
+		vol.HostPath.Path = pathConverter(path)
 
-			if _, err := os.Stat(path); err == nil {
-				continue
-			}
+		if _, err := os.Stat(path); err == nil {
+			_ = os.Chmod(path, permissions)
```
hemanthmantri, line 63: Please add a log.

engine/engine.go
```diff
@@ -52,19 +53,22 @@ func (e *Engine) Setup(ctx context.Context, pipelineConfig *spec.PipelineConfig)
...
 
-			if err := os.MkdirAll(path, 0777); err != nil { //nolint:gomnd
-				return errors.Wrap(err,
-					fmt.Sprintf("failed to create directory for host volume path: %q", path))
-			}
+		if err := os.MkdirAll(path, permissions); err != nil {
+			return errors.Wrap(err,
+				fmt.Sprintf("failed to create directory for host volume path: %q", path))
 		}
+		_ = os.Chmod(path, permissions)
```
hemanthmantri, line 71: Why is this needed?

harshjain12 (approved):

engine/engine.go
anurag-harness: There is no logger here

engine/engine.go
anurag-harness: We still need to give it 0777 if it was created right now

PR #157, opened 2023-06-01
Exec run improvements.

eoinmcafee00:


pipeline/runtime/run_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"reflect"
+	"testing"
+)
+
+func Test_splitCommands(t *testing.T) {
+	tests := []struct {
+		command string
+		want    []string
+	}{
+		{
```
shubham149, line 13: Can you test it with if/else clause, case, function etc. Basically all the possibilities in bash or /bin/sh. Otherwise, it may cause regression

engine/exec/exec.go
```diff
@@ -19,10 +20,10 @@ func Run(ctx context.Context, step *spec.Step, output io.Writer) (*runtime.State
 		return nil, errors.New("step entrypoint cannot be empty")
 	}
 
-	cmdArgs := step.Entrypoint[1:]
-	cmdArgs = append(cmdArgs, step.Command...)
+	// use runner-go shell script logic to generate the script
```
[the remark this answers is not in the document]
tphoney: there would be different behaviour for drone vs cie here. 
with drone we will have already wrapped the commands in a shell script by the time it reaches here. So it would be double wrapped at this point.  a bash script calling another bash script to execute commands.

tphoney (changes requested):

pipeline/runtime/run_test.go
eoinmcafee00: closing this PR. Parsing these possibilities just isn't possible without a full blown bash parser. The change should happen in the platform at the root of the problem, not in the lite-engine as patch work.

PR #158, opened 2023-06-08
add support for lite engine log upload and fix log limit issue

vistaarjuneja:
i) Adds lite engine log upload as part of the destroy call. We have seen many cases where debugging is difficult because of no lite engine logs. This will upload lite-engine.log  to help with debugging. 
ii) Fixes an issue where 5MB log limit was not being honored for VMs.

### DIFF
--- a/handler/destroy.go
+++ b/handler/destroy.go
@@ -23,6 +23,10 @@ import (
 	"github.com/harness/lite-engine/pipeline"
 )
 
+var (
+	liteEngineLogLimit = 2000 // max. number of lines for lite engine logs
+)
+
 func GetLiteEngineLog(osType string) (string, error) {
 	switch osType {
 	// TODO: Add for windows and mac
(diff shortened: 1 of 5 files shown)


shubham149 (approved):

handler/destroy.go
```diff
@@ -5,25 +5,95 @@
...
+			WriteBadRequest(w, err)
+			return
+		}
+
+		if d.LogKey != "" && d.LiteEnginePath != "" {
+			if !d.LogDrone {
+				state := pipeline.GetState()
+				client := state.GetLogStreamClient()
+				logs, uploadErr = GetLiteEngineLog(d.LiteEnginePath)
+				if uploadErr != nil {
```
tphoney, line 71: this isnt an upload error

handler/destroy.go
```diff
@@ -5,25 +5,95 @@
...
+					logger.FromRequest(r).WithField("time", time.Now().
+						Format(time.RFC3339)).WithError(err).Errorln("could not fetch lite engine logs")
+				} else {
```
tphoney, line 74: the else feels redundant. getting the lite engine is ok to fail ?

tphoney (changes requested):

eoinmcafee00 (approved):

handler/destroy.go
vistaarjuneja: added this so that we attempt to upload only if we're able to read the logs

tphoney (approved):

PR #159, opened 2023-06-09
(feat) - strip out set -xe from run commands

eoinmcafee00:


vistaarjuneja:
this should be done on the manager and not here

shubham149:
Customer can also add `set -xe` command as part of their run command. Hence, we should not remove it. Also, if we want to remove set -xe, we probably shouldn't add it in ci-manager.

engine/exec/exec.go
```diff
@@ -21,21 +21,9 @@ func Run(ctx context.Context, step *spec.Step, output io.Writer, isDrone bool) (
 		return nil, errors.New("step entrypoint cannot be empty")
 	}
 
-	// used for printing out the command
-	command := make([]string, len(step.Command))
-	copy(command, step.Command)
-
 	// we shouldn't change this functionality for drone.
```
tphoney, line 24: i would change the comment here, to what was previously on line 24, 
eg `#for CIE we printout the command we are executing`

tphoney (changes requested):

engine/exec/exec.go
```diff
@@ -7,18 +7,24 @@ package exec
...
 
-func Run(ctx context.Context, step *spec.Step, output io.Writer) (*runtime.State, error) {
+func Run(ctx context.Context, step *spec.Step, output io.Writer, isDrone bool) (*runtime.State, error) {
 	if len(step.Entrypoint) == 0 {
 		return nil, errors.New("step entrypoint cannot be empty")
 	}
 
+	// we shouldn't change this functionality for drone.
+	if !isDrone {
+		fmt.Fprintf(output, "Executing command:\n%s\n\n", strings.TrimSpace(strings.Join(step.Command, " ")))
```
vistaarjuneja, line 26: This should also be done on manager - the command formation for K8s/VMs is all done there. K8s also has this same logic of printing the commands initially before the output

PR #166, opened 2023-07-03
(feat): return aggregated os stats as part of destroy call

vistaarjuneja:


logger/logger.go
```diff
@@ -19,7 +19,7 @@ var L = logrus.NewEntry(logrus.StandardLogger())
 // WithContext returns a new context with the provided logger. Use in
 // combination with logger.WithField(s) for great effect.
 func WithContext(ctx context.Context, logger *logrus.Entry) context.Context {
-	return context.WithValue(ctx, loggerKey{}, logger)
+	return context.WithValue(ctx, loggerKey{}, &logrus.Entry{})
```
shubham149, line 22: Is this required?

osstats/collector.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (s *StatsCollector) Stats() *spec.OSStats {
+	return s.stats
+}
+
+// downsample cpu and memory to n points using LTTB
+func (s *StatsCollector) Aggregate() {
+	if len(s.stats.MemGraph.Points) > 0 {
+		s.stats.AvgMemUsagePct = s.memPctSum / float64(len(s.stats.MemGraph.Points))
```
shubham149, line 68: (Optional) Average is not a useful stat. We may use percentile instead. e.g. 90 percentile value for memory and cpu.

logger/logger.go
vistaarjuneja: not needed, undone

osstats/collector.go
vistaarjuneja: yeah, makes sense. I'll wait for gathering some data and see what we need to add/improve in terms of data collection in subsequent PRs. will keep this in mind.

shubham149 (approved):

tphoney (commented):
i assume state is only ever called for CIE pipelines. 
But state is created in setup and in destroy, for both cie and drone

vistaarjuneja:
@tphoney this would be invoked for both drone/cie is my understanding so both would collect stats. We won't start using it in the drone flow in the aws runner side changes yet

tphoney (approved):

PR #171, opened 2023-07-24
Append path variable from docker image if available

raghavharness:
Currently when we run a github action step the we calculate a diff of env variables before and after the step execution. That diff of env variables is then applied to all the subsequent steps. We saw a problem with this when a customer ran a github action and the PATH env variable was changed after the step execution and we eventually set that PATH env variables for all the subsequent steps. Due to this if the PATH env is already set for a docker image then it was getting overridden by the one created by github action step. 

This change is first checking whether the PATH env is provided and also checks if the given docker image also have a PATH env set. If that is true then we just append both the PATHS.

For more context: https://harness.atlassian.net/browse/CI-8897

### DIFF
--- a/engine/docker/docker.go
+++ b/engine/docker/docker.go
@@ -220,9 +220,12 @@ func (e *Docker) create(ctx context.Context, pipelineConfig *spec.PipelineConfig
 		)
 	}
 
-	if err := e.pullDockerImage(ctx, step, pullopts, output); err != nil {
-		return err
+	pulled, pullErr := e.pullDockerImage(ctx, step, pullopts, output)
+	if pullErr != nil {
+		return pullErr
 	}
+
+	e.dockerEnvOverride(ctx, step, pullopts, output, pulled)
 
 	_, err := e.client.ContainerCreate(ctx,
 		toConfig(pipelineConfig, step),
@@ -425,7 +428,7 @@ func (e *Docker) createNetworkWithRetries(ctx context.Context,
 	return err
 }
 
-func (e *Docker) pullDockerImage(ctx context.Context, step *spec.Step, pullOpts types.ImagePullOptions, output io.Writer) error {
+func (e *Docker) pullDockerImage(ctx context.Context, step *spec.Step, pullOpts types.ImagePullOptions, output io.Writer) (bool, error) {
 	pulled := false
 
 	// automatically pull the latest version of the image if requested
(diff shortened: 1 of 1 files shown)


vistaarjuneja:
@raghavharness could you give some more context about this change?

raghavharness:
> @raghavharness could you give some more context about this change and why it's required?

@vistaarjuneja Added in description

vistaarjuneja:
is there a better place we can put this logic? It seems a bit specific and doesn't belong here IMO. The lite engine should just be taking in args and performing some action. Could we do this check in the aws runner instead and update the PATH variable there if the docker image contains PATH? Adding @shubham149 as I might be missing some context

raghavharness:
> is there a better place we can put this logic? It seems a bit specific and doesn't belong here IMO. The lite engine should just be taking in args and performing some action. Could we do this check in the aws runner instead and update the PATH variable there if the docker image contains PATH? Adding @shubham149 as I might be missing some context

@vistaarjuneja Since lite engine is running on same the same VM it can connect to the docker daemon but if we put it on runner let's say then we would not be able to call the docker api's since it's a diff machine right ?

engine/docker/docker.go
```diff
@@ -427,3 +424,71 @@ func (e *Docker) createNetworkWithRetries(ctx context.Context,
...
+		}
+		pulled = true
+	}
+
+	if path, ok := step.Envs[pathEnvVariable]; ok {
+		// pull the image if not already pulled
+		if !pulled {
+			// check if already present on VM else pull
+			pulled = true
+			if exists := e.imageExistsLocally(ctx, step.Image); !exists {
```
shubham149, line 447: This is not correct. Just add in ci-manager run script this line at the top (according to os arch)
```
export PATH=$PATH:<Path in env>
```

shubham149: remove path env from run step

raghavharness: I tried above but that didn't work out.

> remove path env from run step
@shubham149 We set the envs from github action step in runner (https://github.com/drone-runners/drone-runner-aws/blob/ac86ad01ee1ef42e2af6f8065e35cd40ab0714a5/command/harness/step.go#L57). I believe we need to set this for all the subsequent steps.

shubham149: Path that is exported here https://github.com/drone-runners/drone-runner-aws/blob/ac86ad01ee1ef42e2af6f8065e35cd40ab0714a5/command/harness/step.go#L57 won't be useful for steps with docker container.

raghavharness: @shubham149 I see, so would it make sense to remove this logic for container steps at all ? I was earlier thinking to do this but wasn't sure how it can impact the execution.

raghavharness:
Closing this PR since now we'll not add the GA envs to container steps as they won't be useful.
Created PR for that - https://github.com/drone-runners/drone-runner-aws/pull/327

PR #173, opened 2023-07-31
[IAC-824]: Add IACM nudges to lite-engine

nassergonzalez:
In order to improve the error handling from our IACM plugin, we added in the lite-engine nudges for it. I separated the nudges into their own files because it would not be crazy to think that other modules using lite-engine would want to have also a better error handling and the common file would be cluttered with logs really quick and fast. Also by separating the files, each team can just focus their concern in the log part leaving the common functionality untouched.

The new errors look like:
image
image


allofthesepeople:
This feels overly coupled and error prone. e.g.

* lite-engine now has plugin logic directly in it which will only grow with time. 
* log message parsing means any change to the source now needs to be replicated here also.

I would propose it would be better to:

* make use of stderr for its intended purpose for displaying error messages.
* use that as the pipeline error display message.
* helpful tips to fix are part of the error message (if a convention is required here for parsing to some other system we should establish that).

this way the logic specific to the plugin stays in the plugin and coupling reduced. 

Given this PR I assume the number of plugins affected is limited so propose we work toward that. However, if there is an architectural reason lite-engine can't make use of stderr then making use of a final [FATAL] message and applying the same convention for fix tips would be preferred.

pipeline/runtime/ci_nudge.go
```diff
@@ -0,0 +1,16 @@
+package runtime
+
+import (
+	"errors"
+	"github.com/harness/lite-engine/logstream"
+)
+
+func getCINudges() []logstream.Nudge {
+	return []logstream.Nudge{
+		logstream.NewNudge("[Kk]illed", "Increase memory resources for the step", errors.New("out of memory")),
```
urischeiner, line 10: Can we provide some doc around this?

pipeline/runtime/common.go
```diff
@@ -21,14 +21,13 @@ import (
 
 func getNudges() []logstream.Nudge {
 	// <search-term> <resolution> <error-msg>
-	return []logstream.Nudge{
-		logstream.NewNudge("[Kk]illed", "Increase memory resources for the step", errors.New("out of memory")),
-		logstream.NewNudge(".*git.* SSL certificate problem",
-			"Set sslVerify to false in CI codebase properties", errors.New("SSL certificate error")),
```
urischeiner, line 27: We should not mention CI

pipeline/runtime/iacm_nudge.go
```diff
@@ -0,0 +1,140 @@
+package runtime
+
+import (
+	"errors"
+	"github.com/harness/lite-engine/logstream"
+)
+
+func getIACMNudges() []logstream.Nudge {
+	return []logstream.Nudge{
+		logstream.NewNudge("[ERROR] approval step has been rejected",
```
urischeiner, line 10: This should not look like an error. we can remove the "Check why...". 
Can we add the name of the user who perform the operation?

pipeline/runtime/iacm_nudge.go
```diff
@@ -0,0 +1,140 @@
...
+			"Check why the approval step was rejected",
+			errors.New("the approval step has been rejected")),
+		logstream.NewNudge("[ERROR] Error parsing the init_option block",
```
urischeiner, line 13: @nassergonzalez can you please check with support what's the common way to ask users to contact Harness? we should be consistent with the rest of the platform

pipeline/runtime/iacm_nudge.go
```diff
@@ -0,0 +1,140 @@
...
+			"Contact Harness Engineering",
+			errors.New("there was an error starting the plugin")),
```
urischeiner, line 15: can we provide a description on what caused the error?

pipeline/runtime/iacm_nudge.go
```diff
@@ -0,0 +1,140 @@
...
+			"Check the logs for more information. If the problem persist, contact Harness Engineering",
+			errors.New("error during the execution")),
+		logstream.NewNudge("[ERROR] error sending execution event",
+			"Contact Harness Engineering",
+			errors.New("error during the communication with the IACM server")),
+		logstream.NewNudge("[ERROR] Error initializing Terraform",
+			"Error Initializing terraform. Please check the logs. If the problem persists, contact Harness Engineering",
+			errors.New("error during the init step")),
+		logstream.NewNudge("[ERROR] invalid Terraform plugin operation",
+			"This operation is not supported. If you think that the operation should be supported ad is not please contact Harness Engineering",
```
urischeiner, line 56: instead of this message, please write which operations we support

pipeline/runtime/ci_nudge.go
nassergonzalez: This is a CI problem it's not one of our messages

pipeline/runtime/iacm_nudge.go
nassergonzalez: We do not have that information in the logs at this stage so we can't print the user that rejected the step

pipeline/runtime/iacm_nudge.go
nassergonzalez: This specific error is one of those errors that should not happen nor the customer can do anything to alliviate it. For example this one is that when we start the plugin, one of the objects that we populate from the manager was populated wrongly. This is a manager problem literally that is been propagated to the plugin. And because is a manager problem the customers can't do a lot as this is something that needs to be probably fixed by us changing the code of the manager. This is the kind of errors that we would never expect to see in prod _but_ I think that we need to add just in case it happens and we need to investigate what is going on.

vistaarjuneja:
@allofthesepeople I agree with you. The issue is that for plugins today we don't enforce anything as to where the logs should be written to (stdout, or stderr) so I'm not sure if we can make the stderr logic generic (maybe worth a try?). I remember in the past there were issues with some run step logs  going to stdout and some going to stderr. Maybe we introduce some kind of new variable to output error message to for plugins? Open to ideas if you guys have any better generic way to handle this. @shubham149 @tphoney if you guys have any thoughts.

@nassergonzalez the PR looks good to me if you want to go this route - only thing I would suggest is to make sure the error messages are specific to terraform so that we don't misdiagnose for some other CI pipeline with the same error (maybe add terraform: to all error messages?)

tphoney:
implementation seems fine, and i see the need for clarity in the UI for the end user. 
plugins by their nature only care about exit code, and therefor so does CIE / Drone. 
Logs are only for debugging for most plugins, ie slack fails to connect. It will be an "expert" that will diagnose the issue eg change the credentials. 
i think with IACM that the use case has changed. i never knew "nudges" where in there 😸 . And if that suits your UX then go with it. To solve the dependency lock between errors in the plugin and the lite-engine. Can we import this from the plugin or a shared library, so there is only a single source of truth ?

nassergonzalez:
> h

What I was thinking is to eventually if this becomes a PITA (which it would) try to see how can we do it so we have this as a shared library between the plugin and lite-engine. I will give it some thought when we have some time for reflection

allofthesepeople:
@tphoney / @vistaarjuneja 
So if understanding the system correctly, really the error section of the UI is held for exit codes and really the error messages to debug should just be in the logs. 


@nassergonzalez / @urischeiner
If that is the case, I think the correct way to handle this would be to make the error messages in the logs more prescriptive of how to handle failure rather than duplicate logic here?

PR #175, opened 2023-08-08
[CI-8134] print command before step execution

devkimittal:
…mmands

import ordering

removes logic to print out set -x

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -26,6 +26,7 @@ const (
 	DockerSockUnixPath = "/var/run/docker.sock"
 	DockerSockWinPath  = `\\.\pipe\docker_engine`
 	permissions        = 0777
+	boldYellowColor    = "\u001b[33;1m"
 )
 
 type Engine struct {
@@ -89,7 +90,7 @@ func (e *Engine) Destroy(ctx context.Context) error {
 	return e.docker.Destroy(ctx, cfg)
 }
 
-func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*runtime.State, error) {
+func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer, isDrone bool) (*runtime.State, error) {
 	e.mu.Lock()
 	cfg := e.pipelineConfig
 	e.mu.Unlock()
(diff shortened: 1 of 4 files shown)


engine/exec/exec.go
```diff
@@ -27,6 +32,16 @@ func Run(ctx context.Context, step *spec.Step, output io.Writer) (*runtime.State
 	cmd.Env = toEnv(step.Envs)
 	cmd.Stderr = output
 	cmd.Stdout = output
+	if !isDrone && len(step.Command) > 0 {
+		stepCommand := strings.TrimSpace(strings.Join(step.Command, ""))
+		if stepCommand != "" {
+			printCommand := "Executing the following command(s):\n" + stepCommand
+			lines := strings.Split(printCommand, "\n")
```
eoinmcafee00, line 39: Splitting on '/n' isn't going to work for all cases.

eoinmcafee00:
@devkimittal I'd like to see unit test cases .

engine/exec/exec.go
devkimittal: that's what i thought, but it was working, i tried with putting \n, \\n in the command

devkimittal:
@eoinmcafee00 worked alright, the command
Screenshot 2023-08-09 at 4 10 41 PM
Screenshot 2023-08-09 at 4 10 07 PM
 comes as escaped string.

eoinmcafee00:
> …mmands
> 
> import ordering
> 
> removes logic to print out set -x
Fix the commit message please.

engine/docker/docker.go
```diff
@@ -183,12 +185,23 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
 func (e *Docker) Run(ctx context.Context, pipelineConfig *spec.PipelineConfig, step *spec.Step,
-	output io.Writer) (*runtime.State, error) {
+	output io.Writer, isDrone bool) (*runtime.State, error) {
 	// create the container
 	err := e.create(ctx, pipelineConfig, step, output)
 	if err != nil {
 		return nil, errors.TrimExtraInfo(err)
 	}
+
+	if !isDrone && len(step.Command) > 0 {
```
vistaarjuneja, line 195: could we move this to the caller of Run? Trying to move stuff out of engine/docker as it's pretty simple at the moment

devkimittal: good suggestion will do

devkimittal:
Jira : https://harness.atlassian.net/browse/CI-8134

engine/engine.go
```diff
@@ -122,13 +123,27 @@ func (e *Engine) Run(ctx context.Context, step *spec.Step, output io.Writer) (*r
 		vol.Path = pathConverter(vol.Path)
 	}
 
+	printCommand(step, output, isDrone)
```
tphoney, line 126: i would refactor this a little, as it doesnt follow the look and feel in the rest of this function

ie 
```
if !isDrone && len(step.Command) > 0 {
   printCommand(step, output)
end
```
then remove the conditional from the function

tphoney (commented):
just to make it more readable i added a subtle change

vistaarjuneja (approved):

PR #178, opened 2023-08-21
[CI-9246]: Do background execution for detach step only after pulling image

raghavharness:
Currently we are running the go routine for detach step before pulling the image for docker step. We have seen problems in some pipelines where due to image not being pulled the execution fails for different reasons. Here is one such example - https://harness.atlassian.net/browse/CI-9246. 

Have moved the go routine after pulling and creating the docker container.

### DIFF
--- a/engine/docker/docker.go
+++ b/engine/docker/docker.go
@@ -30,6 +30,7 @@ import (
 	"github.com/drone/runner-go/logger"
 	"github.com/drone/runner-go/pipeline/runtime"
 	"github.com/drone/runner-go/registry/auths"
+	"github.com/harness/lite-engine/logstream"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


engine/docker/docker.go
```diff
@@ -183,24 +184,44 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
+		go func() {
+			ctxBg := context.Background()
+			var cancel context.CancelFunc
+			if deadline, ok := ctx.Deadline(); ok {
+				ctxBg, cancel = context.WithTimeout(ctxBg, time.Until(deadline))
+				defer cancel()
+			}
+			e.startContainer(ctxBg, step.ID, output) //nolint:errcheck
+			if wr, ok := output.(logstream.Writer); ok {
+				wr.Close()
```
vistaarjuneja, line 204: is there a reason we are closing the log stream here now? Why are we closing here for detach and somewhere else for not detach?

engine/docker/docker.go
```diff
@@ -183,24 +184,44 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
 
 // Run runs the pipeline step.
 func (e *Docker) Run(ctx context.Context, pipelineConfig *spec.PipelineConfig, step *spec.Step,
-	output io.Writer) (*runtime.State, error) {
+	output io.Writer, isDrone bool) (*runtime.State, error) {
```
vistaarjuneja, line 187: Do we need `isDrone` here? Ideally by this point the behaviour of both drone and harness should be same because the state of the pipeline, step etc is an intermediate representation.
Do we need to have a different behaviour for drone/harness here?

engine/docker/docker.go
```diff
@@ -183,24 +184,44 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
 	err := e.create(ctx, pipelineConfig, step, output)
 	if err != nil {
 		return nil, errors.TrimExtraInfo(err)
 	}
+	// start the execution in go routine if it's a detach step and not drone
+	if !isDrone && step.Detach {
+		go func() {
+			ctxBg := context.Background()
+			var cancel context.CancelFunc
+			if deadline, ok := ctx.Deadline(); ok {
```
vistaarjuneja, line 198: Could we not use the same context which is passed in? Does that get passed around in other steps as well?

engine/docker/docker.go
raghavharness: We close the stream only when the execution is done right. Here we are closing it in routine after the execution is done.

pipeline/runtime/step_executor.go
```diff
@@ -276,10 +277,13 @@ func (e *StepExecutor) executeStep(r *api.StartStepRequest) (*runtime.State, map
 		result = multierror.Append(result, err)
 	}
 
-	// close the stream. If the session is a remote session, the
-	// full log buffer is uploaded to the remote server.
-	if err = wr.Close(); err != nil {
-		result = multierror.Append(result, err)
+	// if err is not nill or it's not a detach step then always close the stream
+	if err != nil || !r.Detach {
```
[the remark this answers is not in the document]
raghavharness: Yeah so I have added the check `err != nil || !r.Detach` which basically means either error is not nil or it's not a detach step. This will ensure that stream is only closed here for non detach steps. We can't close the stream here for detach step since background step runs in a different routine and that is long running. If we close the stream here we'll lose the logs.

engine/docker/docker.go
raghavharness: Actually I don't want to change the drone behaviour unless needed. Since this only came up for Harness, I added the check.

pipeline/runtime/step_executor.go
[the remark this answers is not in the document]
vistaarjuneja: got it.

vistaarjuneja (approved):

PR #180, opened 2023-09-08
Add support to update step status from LE directly

raghavharness:
This change add support to do step status directly from LE to manager.
Added change to store prev step envs which are required for github action plugins. This same logic is used in dlite runner.

### DIFF
--- a/pipeline/runtime/step_executor.go
+++ b/pipeline/runtime/step_executor.go
@@ -42,7 +42,7 @@ const (
 	NotStarted ExecutionStatus = iota
 	Running
 	Complete
-	defaultStepTimeout = 10 * time.Second // default step timeout
+	defaultStepTimeout = 10 * time.Hour // default step timeout
 )
 
 type StepExecutor struct {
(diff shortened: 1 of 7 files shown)


pipeline/runtime/step_executor.go
```diff
@@ -318,6 +359,37 @@ func (e *StepExecutor) run(ctx context.Context, engine *engine.Engine, r *api.St
...
+		return
+	}
+	delegateClient := delegate.NewFromToken(r.StepStatus.Endpoint, r.StepStatus.AccountID, r.StepStatus.Token, true, "")
+	taskResponse := &client.TaskResponse{
+		Data: json.RawMessage(jsonData),
+		Code: "OK",
+		Type: stepStatusUpdate,
+	}
+	if err = delegateClient.SendStatus(context.Background(), r.StepStatus.DelegateID, r.StepStatus.TaskID, taskResponse); err != nil {
+		logrus.WithField("id", r.ID).Errorln("failed to send step status: ", err)
```
vistaarjuneja, line 387: Nit: use `WithError(err)` everywhere

vistaarjuneja (approved):

pipeline/runtime/step_executor.go
raghavharness: done

vistaarjuneja (approved):

PR #182, opened 2023-09-11
[feat]: [CI-8932]: Fix TI for Manual Executions

anurag-harness:


jamie-harness (approved):

jamie-harness:
If we don’t enter manual anymore, we should consider removing it or adding some warning log. But keeping it like this for now should be ok as well

ti/instrumentation/instrumentation.go
```diff
@@ -32,43 +32,53 @@ func getTestSelection(ctx context.Context, runner TestRunner, config *api.RunTes
...
-		} else {
-			files, err = getChangedFiles(ctx, workspace, "", false, log)
+		return selection
+	}
+	// Push+Manual/PR execution
+	var files []ti.File
+	var err error
+	if IsPushTriggerExecution(tiConfig) {
+		lastSuccessfulCommitID, commitErr := getCommitInfo(ctx, stepID, tiConfig)
+		if err != nil {
```
rutvijmehta-harness, line 42:
```suggestion
		if commitErr != nil {
```

rutvijmehta-harness (approved):

PR #185, opened 2023-09-29
[feat]: [CI-9776]: Fix case for ignoreInstr

anurag-harness:


jamie-harness (approved):

ti/instrumentation/instrumentation.go
```diff
@@ -166,6 +166,7 @@ func GetCmd(ctx context.Context, config *api.RunTestConfig, stepID, workspace st
 	// Ignore instrumentation when it's a manual run or user has unchecked RunOnlySelectedTests option
 	isManual := IsManualExecution(cfg)
 	ignoreInstr := isManual || !config.RunOnlySelectedTests
+	cfg.SetIgnoreInstr(ignoreInstr)
```
hemanthmantri, line 169: So is this going to remove TI support for manual triggers?

ti/instrumentation/instrumentation.go
```diff
@@ -175,26 +176,29 @@ func GetCmd(ctx context.Context, config *api.RunTestConfig, stepID, workspace st
...
+			iniFilePath, err = createConfigFile(runner, config.Packages, config.TestAnnotations, workspace, tmpFilePath, fs, log, useYaml)
+			if err != nil {
+				return "", err
+			}
+		} else {
+			config.PreCommand = fmt.Sprintf("export TI_OUTPUT_PATH=%s\n%s", getCgDir(tmpFilePath), config.PreCommand)
+		}
 	}
 
 	// Test splitting: only when parallelism is enabled
```
hemanthmantri, line 204: Please also test the case where parallelism is enabled but Instrumentation is ignored (Selected to run all tests). Ensure there's a test for it.

hemanthmantri: I see places where we are referencing `ignoreInstr` variable directly instead of using GetIgnoreInstr(). Can we keep them consistent so we don't have multiple places to check for issues?

ti/instrumentation/instrumentation.go
anurag-harness: Nope, TI for manual triggers will stil work

ti/instrumentation/instrumentation.go
anurag-harness: Removed referencing of `ignoreInstr`

anurag-harness: There are tests for the parallelism case

ti/instrumentation/instrumentation.go
anurag-harness: Yes we will not go for test selection if it a manual run (according to our old definition of manual run). Now since TI works for manual runs it will go

PR #193, opened 2023-10-12
(CI-5512) enable tty for docker. for colorized logs

tphoney:


engine/docker/convert.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 // returns a container configuration.
-func toConfig(pipelineConfig *spec.PipelineConfig, step *spec.Step) *container.Config {
+func toConfig(pipelineConfig *spec.PipelineConfig, step *spec.Step, tty bool) *container.Config {
```
vistaarjuneja, line 23: we can use `pipelineConfig` directly here for the tty field?

engine/docker/docker.go
```diff
@@ -254,7 +254,7 @@ func (e *Docker) create(ctx context.Context, pipelineConfig *spec.PipelineConfig
 	}
 
 	_, err := e.client.ContainerCreate(ctx,
-		toConfig(pipelineConfig, step),
+		toConfig(pipelineConfig, step, pipelineConfig.TTY),
```
vistaarjuneja, line 257: could we do `toConfig(pipelineConfig, step)`?

tphoney: i need better 👓

vistaarjuneja (approved):

PR #194, opened 2023-10-16
feat: [CI-9787]: internet connectivity check in healthcheck

devkimittal:


handler/health.go
```diff
@@ -20,6 +22,30 @@ func HandleHealth() http.HandlerFunc {
...
 		}
-		WriteJSON(w, response, http.StatusOK)
+		status := http.StatusOK
+
+		if !checkInternetConnectivity() {
+			response = api.HealthResponse{
+				Version: version,
+				OK:      false,
+			}
+			status = http.StatusInternalServerError
```
vistaarjuneja, line 32: can we return a better error code here? And also, there should be some error message that we can log on the client side.

handler/health.go
```diff
@@ -20,6 +22,30 @@ func HandleHealth() http.HandlerFunc {
...
+		}
+
+		WriteJSON(w, response, status)
+	}
+}
+
+func checkInternetConnectivity() bool {
+	dialer := net.Dialer{
```
vistaarjuneja, line 40: I am wary about this change.
What if VMs are coming up but the network is slow. In that case, connectivity check will always fail here and previously passing health checks will now start to fail.

handler/health.go
```diff
@@ -20,6 +22,30 @@ func HandleHealth() http.HandlerFunc {
...
+		Timeout: 1 * time.Second,
+	}
+	conn, err := dialer.Dial("tcp", "8.8.8.8:53")
+
+	if err != nil {
+		logrus.Infoln("Error connecting to 8.8.8.8:53", err)
+		return false
 	}
+	defer conn.Close()
```
raghavharness, line 49: Add this on line 44.

handler/health.go
devkimittal: sure will add a new field message in  HealthResponse

handler/health.go
devkimittal: we can increase the timeout to 2 seconds.

cli/client/http.go
```diff
@@ -193,6 +193,10 @@ func (c *HTTPClient) GetStepLogOutput(ctx context.Context, in *api.StreamOutputR
 
 func (c *HTTPClient) Health(ctx context.Context) (*api.HealthResponse, error) {
 	path := "healthz"
+	if IsHosted(ctx) {
+		path += "?hosted=true"
```
vistaarjuneja, line 197: Use something more generic like `perform_dns_lookup`

cli/client/util.go
```diff
@@ -0,0 +1,17 @@
...
+
+import "context"
+
+type ContextKey string
+
+const (
+	Hosted = ContextKey("hosted")
+)
+
+func IsHosted(ctx context.Context) bool {
```
vistaarjuneja, line 11: Remove this file - it unnecessarily complicates things

cli/client/http.go
```diff
@@ -193,6 +193,10 @@ func (c *HTTPClient) GetStepLogOutput(ctx context.Context, in *api.StreamOutputR
 
 func (c *HTTPClient) Health(ctx context.Context) (*api.HealthResponse, error) {
 	path := "healthz"
+	if IsHosted(ctx) {
```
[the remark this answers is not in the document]
devkimittal: not adding anythign to context here. It was added in an earlier pr and we are just reusing that

cli/client/util.go
devkimittal: we need as a common file  between runner and lite engine

cli/client/http.go
```diff
@@ -193,6 +193,10 @@ func (c *HTTPClient) GetStepLogOutput(ctx context.Context, in *api.StreamOutputR
 
 func (c *HTTPClient) Health(ctx context.Context) (*api.HealthResponse, error) {
```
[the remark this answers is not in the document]
devkimittal: in any case we have to pick it up from context, why not pick here only (requires less change).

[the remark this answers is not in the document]
vistaarjuneja: It should be easy to use this library for a client to interact with a server. Let's say a client wants to use `client/http.go` and wants to do a health check. From the definition, it's not clear what the options are and adding it to context is not intuitive.

cli/client/http.go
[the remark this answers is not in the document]
vistaarjuneja: You can move this to where you call the Health function.
```
  resp, err := client.Health(ctx, (IsHosted(ctx) && isWindows))
```

cli/client/http.go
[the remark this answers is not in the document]
devkimittal: done

cli/client/util.go
devkimittal: removed

vistaarjuneja (approved):

PR #206, opened 2023-11-02
Use default values for python and ruby test generation

jamie-harness:
- added output command to rspec and python runners if user does not provided the command
- added default path to python and ruby if not exist


ti/instrumentation/ruby/rspec.go
```diff
@@ -67,6 +67,10 @@ func (m *rspecRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
 		}
 	}
 	// Run all the tests
+	if userArgs == "" {
+		userArgs = "--format RspecJunitFormatter --out harness_test_results.xml"
```
hemanthmantri, line 71: Shouldn't it be `--out harness_test_results-$HARNESS_NODE_INDEX.xml`?

PR #207, opened 2023-11-03
refactor default test paths for python and ruby TI

jamie-harness:


ti/instrumentation/instrumentation.go
```diff
@@ -219,3 +220,14 @@ func GetCmd(ctx context.Context, config *api.RunTestConfig, stepID, workspace st
 	command := fmt.Sprintf("%s\n%s\n%s", config.PreCommand, testCmd, config.PostCommand)
 	return command, nil
 }
+
+func InjectReportInformation(r *api.StartStepRequest) {
```
rutvijmehta-harness, line 224: Add a comment on why this is required.

ti/instrumentation/utils.go
```diff
@@ -39,10 +39,11 @@ const (
...
 	// Parallelism environment variables
-	harnessStepIndex  = "HARNESS_STEP_INDEX"
-	harnessStepTotal  = "HARNESS_STEP_TOTAL"
-	harnessStageIndex = "HARNESS_STAGE_INDEX"
-	harnessStageTotal = "HARNESS_STAGE_TOTAL"
+	harnessStepIndex    = "HARNESS_STEP_INDEX"
+	harnessStepTotal    = "HARNESS_STEP_TOTAL"
+	harnessStageIndex   = "HARNESS_STAGE_INDEX"
+	harnessStageTotal   = "HARNESS_STAGE_TOTAL"
+	HARNESS_REPORT_PATH = "harness_test_results.xml"
```
rutvijmehta-harness, line 46: Keep the constant in only place and import it everywhere.

ti/instrumentation/common/helper.go
```diff
@@ -5,6 +5,10 @@ import (
 	"github.com/mattn/go-zglob"
 )
 
+const (
+	HARNESS_REPORT_PATH = "harness_test_results.xml"
```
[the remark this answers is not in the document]
rutvijmehta-harness: `HarnessDefaultReportPath` if you want to make it public.

ti/instrumentation/utils.go
jamie-harness: this is removed. Forgot to commit

pipeline/runtime/runtest.go
```diff
@@ -39,6 +39,7 @@ func executeRunTestStep(ctx context.Context, engine *engine.Engine, r *api.Start
 
 	start := time.Now()
 	cmd, err := instrumentation.GetCmd(ctx, &r.RunTest, r.Name, r.WorkingDir, log, r.Envs, tiConfig)
+	instrumentation.InjectReportInformation(r)
```
rutvijmehta-harness, line 42: nit: Add before `GetCmd` or after error check

ti/instrumentation/utils.go
rutvijmehta-harness: It's still here

ti/instrumentation/instrumentation.go
```diff
@@ -219,3 +220,15 @@ func GetCmd(ctx context.Context, config *api.RunTestConfig, stepID, workspace st
...
 	return command, nil
 }
+
+// InjectReportInformation add default test paths information to ruby and python when test runner is invoked without a value
+// This serves as a default
+func InjectReportInformation(r *api.StartStepRequest) {
+	switch strings.ToLower(r.RunTest.Language) {
+	case "ruby", "python":
+		if r.RunTest.Args == "" && len(r.TestReport.Junit.Paths) == 0 {
+			r.TestReport.Junit.Paths = []string{fmt.Sprintf("**/%s*", common.HarnessDefaultReportPath)}
```
rutvijmehta-harness, line 230: How would you handle RunTests with parallelism? Are they writing to a separate folder or separate files?

The above regex won't handle `harness_test_results_0.xml` and your runner is not handling writing to a separate directory for each step.

jamie-harness: for python, it will append to the end (harness_test_results.xml0) in the agent. Ruby still TBD. 

This should only matter for step level. I can add the concurrent index later.

rutvijmehta-harness: Shouldn't it be `harness_test_results_0.xml`

ti/instrumentation/python/unittest.go
```diff
@@ -52,7 +53,7 @@ func (m *unittestRunner) ReadPackages(workspace string, files []ti.File) []ti.Fi
 func (m *unittestRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userArgs, workspace,
 	agentConfigPath, agentInstallDir string, ignoreInstr, runAll bool) (string, error) {
 	if userArgs == "" {
-		userArgs = "--junitxml=harness_test_results.xml -o junit_family='xunit1'"
+		userArgs = fmt.Sprintf("--junitxml='%s${HARNESS_NODE_INDEX}' -o junit_family='xunit1'", common.HarnessDefaultReportPath)
```
rutvijmehta-harness, line 56: You have the value here. Just use it instead of relying on the script.
```suggestion
		userArgs = fmt.Sprintf("--junitxml='%s' -o junit_family='xunit1'", common.HarnessDefaultReportPath, getHarnessDefaultXml())
```

ti/instrumentation/python/pytest.go
```diff
@@ -55,7 +56,7 @@ func (m *pytestRunner) ReadPackages(workspace string, files []ti.File) []ti.File
 func (m *pytestRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userArgs, workspace,
 	agentConfigPath, agentInstallDir string, ignoreInstr, runAll bool) (string, error) {
 	if userArgs == "" {
-		userArgs = "--junitxml=harness_test_results.xml -o junit_family='xunit1'"
+		userArgs = fmt.Sprintf("--junitxml='%s${HARNESS_NODE_INDEX}' -o junit_family='xunit1'", common.HarnessDefaultReportPath)
```
rutvijmehta-harness, line 59: Same as `unittest`

ti/instrumentation/ruby/rspec.go
```diff
@@ -68,7 +68,7 @@ func (m *rspecRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
 	}
 	// Run all the tests
 	if userArgs == "" {
-		userArgs = "--format RspecJunitFormatter --out harness_test_results.xml"
+		userArgs = fmt.Sprintf("--format RspecJunitFormatter --out %s${HARNESS_NODE_INDEX}", common.HarnessDefaultReportPath)
```
rutvijmehta-harness, line 71: Same as `unittest`

ti/instrumentation/instrumentation.go
jamie-harness: index at the end will be easier to handle. We are suggesting the same to the customer for the parallelism. The report does not need to end with xml

PR #209, opened 2023-11-16
[feat]: [CI-9899]: Parse filename in cypress juint xml

anurag-harness:
Please check the xml file in the test example to see the kind of xml we are trying to get the filename for. 
If there is a test suite called Root Suite then the filename used in that must be propagated to other test suites. If any test suite and test case does not have a filename then the one from the Root Suite must be used.

### DIFF
--- a/ti/report/parser/junit/gojunit/ingest.go
+++ b/ti/report/parser/junit/gojunit/ingest.go
@@ -49,15 +49,15 @@ func ingestSuite(root xmlNode, parentFilename string) Suite { //nolint:gocritic
 		Properties: root.Attrs,
 	}
 
-	file := getFilename(root.Attr("file"), parentFilename)
+	parentFilename = getFilename(root.Attr("file"), parentFilename)
 
 	for _, node := range root.Nodes {
 		switch node.XMLName.Local {
 		case "testsuite":
-			testsuite := ingestSuite(node, file)
+			testsuite := ingestSuite(node, parentFilename)
 			suite.Suites = append(suite.Suites, testsuite)
 		case "testcase":
-			testcase := ingestTestcase(node, file)
+			testcase := ingestTestcase(node, parentFilename)
 			suite.Tests = append(suite.Tests, testcase)
 		case "properties":
 			props := ingestProperties(node)
--- a/ti/report/parser/junit/gojunit/ingest.go
+++ b/ti/report/parser/junit/gojunit/ingest.go
@@ -16,20 +16,28 @@ import (
 	ti "github.com/harness/ti-client/types"
 )
 
-const cypressRootSuiteName = "Root Suite"
+const defaultRootSuiteName = "Root Suite"
+
+func getRootSuiteName(envs map[string]string) string {
+	if val, ok := envs["HARNESS_JUNIT_ROOT_SUITE_NAME"]; ok {
+		return val
+	}
+	return defaultRootSuiteName
+}
 
 // findSuites performs a depth-first search through the XML document, and
 // attempts to ingest any "testsuite" tags that are encountered.
-func findSuites(nodes []xmlNode, suites chan Suite, parentFilename string) {
+func findSuites(nodes []xmlNode, suites chan Suite, parentFilename string, envs map[string]string) {
+	rootSuiteName := getRootSuiteName(envs)
 	for _, node := range nodes {
 		switch node.XMLName.Local {
 		case "testsuite":
 			suites <- ingestSuite(node, parentFilename)
-			if node.Attr("name") == cypressRootSuiteName {
+			if node.Attr("name") == rootSuiteName {
 				parentFilename = node.Attr("file")
 			}
 		default:
-			findSuites(node.Nodes, suites, parentFilename)
+			findSuites(node.Nodes, suites, parentFilename, envs)
 		}
 	}
 }
--- a/ti/report/parser/junit/gojunit/ingest.go
+++ b/ti/report/parser/junit/gojunit/ingest.go
@@ -16,33 +16,40 @@ import (
 	ti "github.com/harness/ti-client/types"
 )
 
+const cypressRootSuiteName = "Root Suite"
+
 // findSuites performs a depth-first search through the XML document, and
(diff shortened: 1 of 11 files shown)


ti/report/parser/junit/gojunit/ingest.go
```diff
@@ -16,33 +16,40 @@ import (
...
 // findSuites performs a depth-first search through the XML document, and
 // attempts to ingest any "testsuite" tags that are encountered.
-func findSuites(nodes []xmlNode, suites chan Suite) {
+func findSuites(nodes []xmlNode, suites chan Suite, parentFilename string) {
 	for _, node := range nodes {
 		switch node.XMLName.Local {
 		case "testsuite":
-			suites <- ingestSuite(node)
+			suites <- ingestSuite(node, parentFilename)
+			if node.Attr("name") == cypressRootSuiteName {
```
rutvijmehta-harness, line 28: Come up with a better name and don't make it cypress specific
```suggestion
			if node.Attr("name") == someBetterName {
```

ti/report/parser/junit/gojunit/ingest.go
```diff
@@ -16,33 +16,40 @@ import (
 	ti "github.com/harness/ti-client/types"
 )
 
+const cypressRootSuiteName = "Root Suite"
```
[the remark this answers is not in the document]
anurag-harness: Can we keep this as the default incase nothing is provided by the user?

ti/report/parser/junit/gojunit/ingest.go
```diff
@@ -16,33 +16,40 @@ import (
...
 
-func ingestSuite(root xmlNode) Suite { //nolint:gocritic
+func ingestSuite(root xmlNode, parentFilename string) Suite { //nolint:gocritic
 	suite := Suite{
 		Name:       root.Attr("name"),
 		Package:    root.Attr("package"),
 		Properties: root.Attrs,
 	}
 
+	file := getFilename(root.Attr("file"), parentFilename)
```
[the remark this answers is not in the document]
anurag-harness: Why should this be changed to parentFilename? It should be file right since this is what we will now consider as the file for this Test Suite

ti/report/parser/junit/gojunit/ingest.go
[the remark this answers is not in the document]
anurag-harness: When I take the value from an env variable should I do os.Lookup in this file or should I use the config struct and pass that throughout the code to this point?

ti/report/parser/junit/gojunit/ingest.go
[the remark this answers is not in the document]
rutvijmehta-harness: It's just naming convention. It's technically a `parent` since you're passing it to `ingestSuite`. You can just re-use the same variable name since it's a recursion.

ti/report/parser/junit/gojunit/ingest.go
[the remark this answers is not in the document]
rutvijmehta-harness: `os.Lookup` won't work. You'll have to get the env variable value via method arg.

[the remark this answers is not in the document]
anurag-harness: done

ti/report/parser/junit/gojunit/ingest.go
anurag-harness: Done

ti/report/parser/junit/gojunit/ingest.go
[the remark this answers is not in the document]
anurag-harness: Done

ti/report/parser/junit/gojunit/ingest.go
```diff
@@ -16,33 +16,48 @@ import (
...
+	if val, ok := envs["HARNESS_JUNIT_ROOT_SUITE_NAME"]; ok {
+		return val
+	}
+	return defaultRootSuiteName
+}
+
 // findSuites performs a depth-first search through the XML document, and
 // attempts to ingest any "testsuite" tags that are encountered.
-func findSuites(nodes []xmlNode, suites chan Suite) {
+func findSuites(nodes []xmlNode, suites chan Suite, parentFilename string, envs map[string]string) {
```
rutvijmehta-harness, line 30: You don't need to pass the full map. Just pass what you need (`parentFilename`)

ti/report/parser/junit/gojunit/ingest.go
```diff
@@ -16,33 +16,48 @@ import (
 	ti "github.com/harness/ti-client/types"
 )
 
+const defaultRootSuiteName = "Root Suite"
+
+func getRootSuiteName(envs map[string]string) string {
+	if val, ok := envs["HARNESS_JUNIT_ROOT_SUITE_NAME"]; ok {
+		return val
+	}
+	return defaultRootSuiteName
```
rutvijmehta-harness, line 25: I still prefer to not have `defaultRootSuiteName`. You can merge if the team agrees.

ti/report/parser/junit/gojunit/ingest.go
anurag-harness: I need to pass rootSuiteName as well right?

anurag-harness: Or should I use a global variable for it?

ti/report/parser/junit/junit.go
```diff
@@ -32,7 +32,7 @@ func ParseTests(paths []string, log *logrus.Logger) []*ti.TestCase {
 	totalTests := 0
 	var tests []*ti.TestCase
 	for _, file := range files {
-		suites, err := gojunit.IngestFile(file)
+		suites, err := gojunit.IngestFile(file, envs)
```
rutvijmehta-harness, line 35: Pass the parent filename here

ti/report/parser/junit/gojunit/ingest.go
anurag-harness: Discussed on call

rutvijmehta-harness (approved):

PR #210, opened 2023-11-17
Prepend to rspec helper file instead of append

jamie-harness:


tphoney (commented):
normally there is some strings at the start of the helper  files, i would add the line where they have requires.

PR #216, opened 2023-12-18
[feat]: [CI-9776]: Fix case for ignoreInstr

anurag-harness:
Before if it was a manual execution, RunOnlySelected tests was made false in the test selection part but now that was skipped. And hence in case of manual execution we were running selected tests which were empty since we did not select any tests to run. Found this issue when trying to setup harness code with TI since that is the only flow we are seeing TI defined manual executions still.


Aishwarya-Lad (approved):
LGTM

PR #218, opened 2023-12-19
(CI-10373) fixes for ruby TI

tphoney:


ti/instrumentation/ruby/helper.go
```diff
@@ -83,26 +83,26 @@ func UnzipAndGetTestInfo(agentInstallDir string, log *logrus.Logger) (scriptPath
 	return scriptPath, nil
 }
 
-func WriteGemFile(workspace, repoPath string) error {
-	pattern := fmt.Sprintf("%s/**/Gemfile", workspace)
+func AddHarnessRubyAgentToGemfile(repoPath string, log *logrus.Logger) error {
+	cmd := exec.Command("bundle", "add", "harness_ruby_agent", "--path", fmt.Sprintf("%q", repoPath), "--version", "~> 0.0.1") // #nosec
```
hemanthmantri, line 87: This seems to assume bundle is installed on the VM/POD and is in the $PATH, right?

hemanthmantri: Will this also install the gem or just add to the Gem file?

jamie-harness: Does this fail the build if pwd is not a ruby repo? Should we document user need cd into the corresponding folder that contains the ruby stuff in pre-command?

ti/instrumentation/ruby/rspec.go
```diff
@@ -56,9 +56,13 @@ func (m *rspecRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
 		if err != nil {
 			return "", err
 		}
-		err = WriteGemFile(workspace, repoPath)
+		err = AddHarnessRubyAgentToGemfile(repoPath, m.log)
 		if err != nil {
```
jamie-harness, line 60: we might want to ignore this error.Print a warning and move forward

ti/instrumentation/ruby/rspec.go
```diff
@@ -56,9 +56,13 @@ func (m *rspecRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
...
-			return testCmd, err
+			return "", err
+		}
+		err = AddRspecJunitFormatterToGemfile(repoPath, m.log)
+		if err != nil {
```
jamie-harness, line 64: same as above

ti/instrumentation/ruby/rspec.go
```diff
@@ -56,9 +56,13 @@ func (m *rspecRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
 		if err != nil {
 			return "", err
 		}
-		err = WriteGemFile(workspace, repoPath)
+		err = AddHarnessRubyAgentToGemfile(repoPath, m.log)
```
jamie-harness, line 59: we might want to ignore this error.Print a warning and move forward

PR #229, opened 2024-01-08
Treat all files in the package as changed if build.bazel file changes.

Aishwarya-Lad:
Add the package level test files to changed file list when a build.bazel file changes to run TI on.
 originally TI would select entire test suite when any build.bazel file is changed


ti/instrumentation/utils.go
```diff
@@ -264,6 +266,44 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+func addBazelFilesToChangedFiles(ctx context.Context, workspace string, log *logrus.Logger, res []ti.File) ([]ti.File, error) {
+	//resCopy := copy(resCopy, res)
+	changedRes := []ti.File{}
+	//copy(copiedRes, res)
+
+	for _, file := range res {
+		log.Infoln("A, currently processing file : ", file)
+		if strings.HasSuffix(string(file.Name), "BUILD.bazel") {
+			fileName := "//" + strings.Replace(file.Name, "/BUILD.bazel", "", -1)
+			c := fmt.Sprintf("cd %s; %s query 'kind(\"java_library\", %s:tests)' --output=build "+
```
hemanthmantri, line 279: Can you add some comments explaining what this long command does? Also, you are looking at `java_library` only. Is that the only target/rule to look for?

Aishwarya-Lad: Modified to include rules beginning with "java", so java_library and java_binary

ti/instrumentation/instrumentation.go
```diff
@@ -29,6 +29,7 @@ const (
 func getTestSelection(ctx context.Context, runner TestRunner, config *api.RunTestConfig, fs filesystem.FileSystem,
 	stepID, workspace string, log *logrus.Logger, isManual bool, tiConfig *tiCfg.Cfg) ti.SelectTestsResp {
 	selection := ti.SelectTestsResp{}
+	log.Infoln("Aishwarya - testing lite engine VM setup.")
```
anurag-harness, line 32: Remove this

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,108 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+
+	var changedRes []ti.File
+	for _, file := range res {
+		if strings.HasSuffix(string(file.Name), "BUILD.bazel") {
+			//If BUILD.Bazel at the root of harness-core is modified, run all tests /... by keeping BUILD.Bazel in changed files list
+			if len(strings.Replace(file.Name, "BUILD.bazel", "", -1)) == 0 {
+				return res, nil
+			}
+			bazelQueryInput :=  strings.Replace(file.Name, "/BUILD.bazel", "", -1)
+			directory := workspace + "/" + strings.Replace(file.Name, "/BUILD.bazel", "", -1)
```
anurag-harness, line 299: You can use filpath.Join instead of using + to get the directory name

anurag-harness (changes requested):
Update the go.mod and go.sum with the new ti-client you are using
https://github.com/harness/ti-client/pull/16/files

ti/instrumentation/instrumentation.go
```diff
@@ -43,28 +46,43 @@ func getTestSelection(ctx context.Context, runner TestRunner, config *api.RunTes
...
+	//check ticonfig params to allow bazel optimization, and get threshold for max file count
+	tiConfigYaml, err := getTiConfig(workspace, fs)
+	if err != nil {
+		log.Infoln("Ti config parsing before selectTests fails")
+	}
+
+	//skip bazel src inspection if optimization in config not selected
+	if tiConfigYaml.Config.BazelOptimization {
+		//add src files listed in java target rules in the BUILD.bazel files if BUILD.bazel is changed
+		files, moduleList, err = addBazelFilesToChangedFiles(ctx, workspace, log, files, tiConfigYaml.Config.BazelFileCount)
```
rutvijmehta-harness, line 81: What happens if `err != nil`?

ti/instrumentation/instrumentation.go
```diff
@@ -212,6 +230,20 @@ func GetCmd(ctx context.Context, config *api.RunTestConfig, stepID, workspace st
 	}
 
 	testCmd, err := runner.GetCmd(ctx, selection.Tests, config.Args, workspace, iniFilePath, artifactDir, cfg.GetIgnoreInstr(), !config.RunOnlySelectedTests)
+
+	if len(modules) != 0 && strings.Contains(testCmd, "bazel") {
```
rutvijmehta-harness: Move this whole logic to `GetCmd` of bazel.go
```suggestion
	if tiConfigYaml.Config.BazelOptimization & config.BuildTool == "bazel" {
	   testCmd = getBazelTestCmd(...)
	}
```

ti/instrumentation/instrumentation.go
```diff
@@ -224,6 +256,16 @@ func GetCmd(ctx context.Context, config *api.RunTestConfig, stepID, workspace st
 	return command, nil
 }
 
+// get modules to add as test target to the bazel test command
```
rutvijmehta-harness, line 258: Move this to bazel.go as well

ti/instrumentation/instrumentation.go
```diff
@@ -43,28 +46,43 @@ func getTestSelection(ctx context.Context, runner TestRunner, config *api.RunTes
...
 		files, err = getChangedFilesPR(ctx, workspace, log)
 		if err != nil || len(files) == 0 {
 			log.Errorln("Unable to get changed files list for PR. Running all the tests.", "error", err)
 			config.RunOnlySelectedTests = false
-			return selection
+			return selection, moduleList
 		}
 	}
+
+	//check ticonfig params to allow bazel optimization, and get threshold for max file count
```
rutvijmehta-harness, line 72: Move this logic to another method `filterChangedFiles`

ti/instrumentation/instrumentation.go
```diff
@@ -43,28 +46,43 @@ func getTestSelection(ctx context.Context, runner TestRunner, config *api.RunTes
...
+	if err != nil {
+		log.Infoln("Ti config parsing before selectTests fails")
+	}
+
+	//skip bazel src inspection if optimization in config not selected
+	if tiConfigYaml.Config.BazelOptimization {
+		//add src files listed in java target rules in the BUILD.bazel files if BUILD.bazel is changed
+		files, moduleList, err = addBazelFilesToChangedFiles(ctx, workspace, log, files, tiConfigYaml.Config.BazelFileCount)
+		log.Infoln("Changed file list is: ", files)
+		log.Infoln("Changed module list is: ", moduleList)
```
rutvijmehta-harness, line 83: `Changed files after filter:`

ti/instrumentation/utils.go
```diff
@@ -264,6 +266,136 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+		}
+	}
+	return javaFileNames
+}
+
+// takes file name, runs bazel queries extract src globs defined in java rules, viz java_library and java_test
+func getJavaRulesFromBazel(file string, workspace string, ctx context.Context) ([]byte, []byte, []byte, error) {
+	bazelQueryInput := strings.Replace(file, "/BUILD.bazel", "", -1)
+	//query to fetch test src globs from java rule
+	c1 := fmt.Sprintf("cd %s; %s query 'kind(\"java\", %s:*)' --output=build | grep 'srcs =' ", workspace, bazelCmd, bazelQueryInput)
```
rutvijmehta-harness, line 362: Add an example of how this command looks in plain text. Make this change everywhere you're running a query.

ti/instrumentation/utils.go
```diff
@@ -264,6 +266,136 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+	//query to get test kind for module list
+	c3 := fmt.Sprintf("cd %s; %s query 'kind(\"java_library\", %s:tests)' --output=build | grep 'srcs =' ", workspace, bazelCmd, bazelQueryInput)
+	cmdArgs3 := []string{"-c", c3}
+	javaTestKindOutput, err := execCmdCtx(ctx, "sh", cmdArgs3...).Output()
+
+	return javaKindOutput, countQueryOutput, javaTestKindOutput, err
+}
+
+// takes a directory name and adds all java files within that module/package in changed file res
+func getAllJavaFilesInsideDirectory(directory string, res []ti.File, file ti.File, uniqueFiles map[string]struct{}) ([]ti.File, error) {
```
rutvijmehta-harness, line 382: I believe we already have a util for this.

ti/instrumentation/instrumentation.go
```diff
@@ -7,11 +7,11 @@ package instrumentation
 import (
 	"context"
 	"fmt"
+	"github.com/sirupsen/logrus"
```
rutvijmehta-harness, line 10: Fix import order everywhere. Check go docs for best practices.

ti/instrumentation/utils.go
Aishwarya-Lad: do you mean Go utility? It has but I want to just consider .java files not scripts and other files.

ti/instrumentation/instrumentation.go
Aishwarya-Lad: Ideally we should login an error and continue the code-flow (that means bazel optimization logic is not applied)

ti/instrumentation/instrumentation.go
```diff
@@ -211,7 +242,12 @@ func GetCmd(ctx context.Context, config *api.RunTestConfig, stepID, workspace st
 		computeSelectedTests(ctx, config, log, runner, &selection, workspace, envs, cfg)
 	}
 
-	testCmd, err := runner.GetCmd(ctx, selection.Tests, config.Args, workspace, iniFilePath, artifactDir, cfg.GetIgnoreInstr(), !config.RunOnlySelectedTests)
+	// set runnerArg for bazel runner
+	runnerArgs := common.RunnerArgs{}
+	runnerArgs.ModuleList = modules
+
+	testCmd, err := runner.GetCmd(ctx, selection.Tests, config.Args, workspace, iniFilePath, artifactDir, cfg.GetIgnoreInstr(), !config.RunOnlySelectedTests, runnerArgs)
+	log.Infoln("Bazel testCmd is :", testCmd)
```
anurag-harness, line 250: Remove this log line

ti/instrumentation/java/bazel.go
```diff
@@ -236,3 +252,17 @@ func (b *bazelRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
 	}
 	return fmt.Sprintf("%s %s %s %s", bazelCmd, userArgs, instrArg, testList), nil
 }
+
+// parse module name from rule
```
hemanthmantri, line 256: Please add example input and output int the comment.

ti/instrumentation/java/bazel.go
```diff
@@ -236,3 +252,17 @@ func (b *bazelRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
...
+func getModuleFromRule(rule string) string {
+	// parse the rule to extract module
+	splitRule := strings.Split(strings.TrimPrefix(rule, "//"), ":")
+	if strings.Contains(splitRule[0], "/") {
```
hemanthmantri, line 260: Need to check splitRule has 0th index (not empty)

ti/instrumentation/java/bazel.go
```diff
@@ -236,3 +252,17 @@ func (b *bazelRunner) GetCmd(ctx context.Context, tests []ti.RunnableTest, userA
...
+		splitModule := strings.Split(splitRule[0], "/")
```
hemanthmantri, line 261: Ditto.

hemanthmantri: Add Unit Tests for this method to ensure unexpected formats are not going to cause a panic.

ti/instrumentation/instrumentation.go
```diff
@@ -81,7 +88,30 @@ func getTestSelection(ctx context.Context, runner TestRunner, config *api.RunTes
...
+	// check ticonfig params to allow bazel optimization, and get threshold for max file count
+	tiConfigYaml, err := getTiConfig(workspace, fs)
+	if err != nil {
+		log.Infoln("Ti config parsing before selectTests fails")
+	}
+
+	// skip bazel src inspection if optimization in config not selected
+	if tiConfigYaml.Config.BazelOptimization {
+		// add src files listed in java target rules in the BUILD.bazel files if BUILD.bazel is changed
+		newFiles, moduleList, err = addBazelFilesToChangedFiles(ctx, workspace, log, files, tiConfigYaml.Config.BazelFileCount)
```
hemanthmantri, line 107: Can we not read integers from YAML directly? I see you are passing this as a string. The util method should take integer and callers need to make sure the input is passed correctly after validation. If validation fails, we should not panic/fail, just ignore optimization.

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
 	return res, nil
 }
 
+// addBazelFilesToChangedFiles takes a list of files and removes bazel build files and adds java files listed in target src globs
+func addBazelFilesToChangedFiles(ctx context.Context, workspace string, log *logrus.Logger, res []ti.File, bazelFileCount string) ([]ti.File, []string, error) {
```
hemanthmantri, line 271: We should name the parameters such that they are self explanatory for the most part. For example "res" is not.
Also consider `bazelFileCount` -> `bazelFileCountThreshold`

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+	var moduleList []string
+	threshold, err := strconv.Atoi(bazelFileCount)
+	if err != nil {
+		return res, moduleList, fmt.Errorf("bazelFileCount not set correctly in ticonfig.yml, expecting number found character %v ", bazelFileCount)
```
hemanthmantri, line 275: This is a client call validation not the method's (since you chose to make it a util method).

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+	}
+
+	// map to prevent duplicate files
+	uniqueFiles := make(map[string]struct{})
+
+	var changedRes []ti.File
+	for _, file := range res {
+		if !strings.HasSuffix(file.Name, "BUILD.bazel") {
```
hemanthmantri, line 283: Bazel convention is `BUILD` but Harness seems to use `BUILD.bazel` so please add a comment here that this needs to change when we support this feature more generically for bazel. Also make sure we are not using `BUILD` anywhere in harness-core

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+			// add non BUILD.bazel files to the changed files result directly and skip remainder
+			if _, exists := uniqueFiles[file.Name]; !exists {
+				changedRes = append(changedRes, file)
+				uniqueFiles[file.Name] = struct{}{}
+			}
+			continue
+		}
+		// If BUILD.Bazel at the root of harness-core is modified, run all tests /... by keeping BUILD.Bazel in changed files list
+		if strings.Replace(file.Name, "BUILD.bazel", "", -1) == "" {
+			log.Infoln("Changed file list is: ", res)
```
hemanthmantri, line 293: The `res` is not changing in this loop but we are printing it every-time. Should be printed once.

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+			log.Infoln("Determined to run all tests /... : ")
+			return res, moduleList, nil
```
hemanthmantri, line 295: At this point moduleList is empty. Is that fine?

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+			continue
+		}
+		count, err := strconv.Atoi(strings.TrimSpace(string(countAllJavaSrcs)))
+		if err != nil {
+			count = 0
+		}
+		directory := filepath.Join(workspace, strings.Replace(file.Name, "/BUILD.bazel", "", -1))
+		// if no srcs present in the bazel build file, then select all files under this directory as changed
+		if count == 0 {
+			message := fmt.Sprintf("No src detected in Bazel %v, selecting all java files as changed files inside the directory %v", file.Name, directory)
```
hemanthmantri, line 314: s/selecting/considering

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+			log.Infoln(message)
+			changedRes, _ = getAllJavaFilesInsideDirectory(directory, changedRes, file, uniqueFiles)
+		} else if count >= threshold {
+			// if count of files changed due to bazel build file modification is over the limit set then update the module list with test target
+			message := fmt.Sprintf("%v Files changed after changing bazel file %v; crosses limit set for bazelFileCount", count, file.Name)
+			log.Infoln(message)
+			if javaTestKindOutput != nil {
+				splitName := strings.Split(file.Name, "/")
+				moduleName := splitName[0]
```
hemanthmantri, line 323: Nil/empty check

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+					changedRes = append(changedRes, ti.File{Status: file.Status, Name: name})
+					uniqueFiles[name] = struct{}{}
+				}
+			}
+		}
+	}
+	return changedRes, moduleList, nil
+}
+
+// takes bazelQueryOutput string and parses it to extract .java files and returns in a list
```
hemanthmantri, line 341: Please add a sample input here.

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+func extractJavaFilesFromQueryOutput(bazelOutput []byte) []string {
+	var javaFileNames []string
+	// Outer Regex to extract only srcs
+	pattern1 := `srcs = \[[^\]]+\]`
```
hemanthmantri, line 345: Does this work for cases where srcs section is split in multiple lines? Also can the `=` sign have no spaces before and after it?

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+	return javaFileNames
+}
+
+// takes file name, runs bazel queries extract src globs defined in java rules, viz java_library and java_test
+func getJavaRulesFromBazel(ctx context.Context, file, workspace string) (allSrcs, countAllSrcs, testSrcs []byte, err error) {
+	bazelQueryInput := strings.Replace(file, "/BUILD.bazel", "", -1)
+
+	// query to fetch test src globs from java rule
+	// eg of the query is: bazel query 'kind("java", 980-commons:*)' --output=build  | grep 'srcs ='
+	c1 := fmt.Sprintf("cd %s; %s query 'kind(\"java\", %s:*)' --output=build | grep 'srcs =' ", workspace, bazelCmd, bazelQueryInput)
```
hemanthmantri, line 369: Please check if the if the space is required before `=`

ti/instrumentation/utils.go
```diff
@@ -264,6 +267,151 @@ func getChangedFiles(ctx context.Context, workspace string, log *logrus.Logger,
...
+	javaKindOutput, err := execCmdCtx(ctx, "sh", cmdArgs1...).Output()
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to run bazel query %v, encountered %v as %v has no java rule", c1, err, file)
+	}
+
+	// query to count the scr files in bazelQueryOutput
+	// eg of the query is: bazel query 'kind("java", 332-ci-manager/app:*)' --output=build  | grep 'srcs =' | grep -o '\.java' | wc -l
+	c2 := fmt.Sprintf("cd %s; %s query 'kind(\"java\", %s:*)' --output=build | grep 'srcs =' | grep -o '\\.java' | wc -l", workspace, bazelCmd, bazelQueryInput)
+	cmdArgs2 := []string{"-c", c2}
+	countQueryOutput, err := execCmdCtx(ctx, "sh", cmdArgs2...).Output()
```
hemanthmantri, line 380: Do these commands have a timeout? We should not hang for a long time and give up after some timeout.

ti/instrumentation/utils.go
hemanthmantri: Please name the params to indicate their purpose. `res` in the input doesn't sound right.

hemanthmantri (commented):
Added some style comments mainly around error handling and naming. We should strive to never cause an uncaught exception in production and so practice defensive coding. An exception can cause the whole workflow to fail - we should fallback to using the unoptimized path on any errors.

jamie-harness:
This is a minor comment from me - I don't have a strong opinion on this but I also don't like having bazel query stuff in non-bazel runner place. Maybe we can pass the changed file to the runner in the new arg you provided and handle it in the runner itself?

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/go-chi/chi/v5 v5.0.8
 	github.com/gofrs/uuid v4.4.0+incompatible
 	github.com/golang/mock v1.6.0
-	github.com/harness/ti-client v0.0.0-20231018000842-d164bcf4f802
+	github.com/harness/ti-client v0.0.0-20240112195129-fc1b3987a83b
```
jamie-harness, line 19: let's merge the client PR and update this before merging this one

go.sum
```diff
@@ -156,6 +158,7 @@ github.com/sirupsen/logrus v1.9.0/go.mod h1:naHLuLoDiP4jHNo9R0sCBMtWGeIprob74mVs
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/objx v0.4.0/go.mod h1:YvHI0jy2hoMjB+UWwv71VJQ9isScKT/TqJzVSSt89Yw=
+github.com/stretchr/objx v0.5.0 h1:1zr/of2m5FGMsad5YfcqgdqdWrIhu+EBEJRhR1U7z/c=
```
jamie-harness, line 161: run go mod tidy to clean this up a little

ti/instrumentation/utils.go
Aishwarya-Lad: It would only be printed once for root BUILD.bazel file and we'll break out of the func.

PR #231, opened 2024-01-15
IACM proposal

nassergonzalez:
Background

Currently in IACM we had a problem where an apply step reached the timeout while the terraform apply command was running. As a result of the current code, the container was killed and the terraform.tfstate was lost. We thought that we could put some signal handling in the plugin to fix this problem but during the investigation on the plugin side, we realised that no SIGTERM/SIGINT was sent.

After checking the code in the destroy.go in lite engine I think that what is happening is:

- We send the kill signal to the pod. Even if the signal could be change, seems that what is been sent right now is a SIGKILL which can be handled.
- During testing in the lite-engine side, even changing the signal to SIGINT was not enough. The problem then was that the ContainerRemove deleted the pod even if the state of the container was still running. That caused the SIGINT never been caught or handled in the plugin.

Proposal

Use ContainerStop instead of ContainerKill. This sends SIGINT and allows a grace period for shutdown. With this, plugins have 30 seconds to be able to handle the signal, and save their state if necessary. Docker will kill the container after 30 seconds so there should be no risks of containers living for more than that time

Avoid executing the ContainerRemove until all the containers have an exited status. This way we honour the 30 seconds and don't remove the containers until the containers have exited

### DIFF
--- a/engine/docker/docker.go
+++ b/engine/docker/docker.go
@@ -157,10 +157,31 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
 	containers := e.containers
 	e.mu.Unlock()
 
+	timeout := time.Second * 30 // similar to k8s default timeout in pods
 	// stop all containers
 	for _, ctrName := range containers {
-		if err := e.client.ContainerKill(ctx, ctrName, "9"); err != nil {
-			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to kill container")
+		if err := e.client.ContainerStop(ctx, ctrName, &timeout); err != nil {
+			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to stop the container")
+		}
+
+		// Before removing the container we want to be sure that it's in a healthy state to be removed.
+
+		containerStatus, err := e.client.ContainerInspect(ctx, ctrName)
+
+		if err != nil {
+			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to retrieve container stats")
+		}
+
+		for {
+			if containerStatus.State.Status == "removing" || containerStatus.State.Status == "running" {
+				time.Sleep(1 * time.Second)
+				containerStatus, err = e.client.ContainerInspect(ctx, ctrName)
+				if err != nil {
+					logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to retrieve container stats")
+				}
+			} else {
+				break
+			}
 		}
 	}
 


engine/docker/docker.go
```diff
@@ -157,10 +157,31 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
 	for _, ctrName := range containers {
-		if err := e.client.ContainerKill(ctx, ctrName, "9"); err != nil {
-			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to kill container")
+		if err := e.client.ContainerStop(ctx, ctrName, &timeout); err != nil {
+			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to stop the container")
+		}
+
+		// Before removing the container we want to be sure that it's in a healthy state to be removed.
+
+		containerStatus, err := e.client.ContainerInspect(ctx, ctrName)
```
scottyw-harness, line 169: Can maybe move these inside the for loop?

i.e. get status, if it is removing or running then sleep 1 and repeat, otherwise we're done.

engine/docker/docker.go
```diff
@@ -157,10 +157,31 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
+	timeout := time.Second * 30 // similar to k8s default timeout in pods
 	// stop all containers
 	for _, ctrName := range containers {
-		if err := e.client.ContainerKill(ctx, ctrName, "9"); err != nil {
-			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to kill container")
+		if err := e.client.ContainerStop(ctx, ctrName, &timeout); err != nil {
+			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to stop the container")
+		}
+
+		// Before removing the container we want to be sure that it's in a healthy state to be removed.
```
scottyw-harness, line 167: I think we should have 2 for loops: one for loop to stop all containers, one for loop to check status and kill those that didn't stop.

That would parallelize stop.

engine/docker/docker.go
```diff
@@ -157,10 +157,31 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
+
+		containerStatus, err := e.client.ContainerInspect(ctx, ctrName)
+
+		if err != nil {
+			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to retrieve container stats")
+		}
+
+		for {
+			if containerStatus.State.Status == "removing" || containerStatus.State.Status == "running" {
+				time.Sleep(1 * time.Second)
```
scottyw-harness, line 177: We need some form of kill in here e.g. wait 30s for it to stop properly and kill it if it doesn't.

scottyw-harness (approved):

engine/docker/docker.go
nassergonzalez: When using the ContainerStop, docker will kill the container for us if after 30 seconds the container has not been stopped so there should be no need for a kill

engine/docker/docker.go
nassergonzalez: The stop should stop the containers and if after 30 seconds the containers are not stopped, docker will kill them for us so i do not think that we need a second for loop

engine/docker/docker.go
nassergonzalez: I was aiming for a do while. For a lot of containers, the state will be existed so we jump from 169 to 183 directly. For example, imagine a IACM pipeline, by the time the apply is running, the init and plan container already stopped (and exited) so when those containers are been inspected here, the state will be already exited

engine/docker/docker.go
```diff
@@ -157,10 +157,31 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
+		// Before removing the container we want to be sure that it's in a healthy state to be removed.
+
+		containerStatus, err := e.client.ContainerInspect(ctx, ctrName)
+
+		if err != nil {
+			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to retrieve container stats")
+		}
+
+		for {
+			if containerStatus.State.Status == "removing" || containerStatus.State.Status == "running" {
```
vistaarjuneja, line 176: if L169 returns an error, should we be going in this loop?

engine/docker/docker.go
```diff
@@ -157,10 +157,31 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
 	containers := e.containers
 	e.mu.Unlock()
 
+	timeout := time.Second * 30 // similar to k8s default timeout in pods
```
vistaarjuneja, line 160: I would be cautious with introducing this for all flows. This lite-engine is currently used in our hosted builds and the destroy flow there does not need to be graceful. If we are changing that behaviour, could we add a flag in the pipeline config which could dictate whether it's a graceful destroy or a force destroy (and it could be force by default and graceful for IACM?). That would make this change a lot safer and it would get some soak time IMO.

nassergonzalez: So I do agree with you but it means that we probably need to expand the API and that is where @tphoney or you can help. We could 
- New method, instead of Destroy, Stop. Then we have the logic there and people could use destroy if they want to, or stop. The problem of this is:
- Because is a new API we probably need something to determine if we need to use Destroy or Stop. I guess that is a change in the runners?

or

- We could expand the Destroy. Maybe spec.PipelineConfig can have a new field, something around:
- - Destroy bool -> if true use destroy path if not, follow the stop path

Any suggestion are really appreciated and I am happy to change/implement them but I would need guidance of what needs to be done and changed

engine/docker/docker.go
vistaarjuneja, line 169: Curious - would we need this? Could the rest of the code remain as it is? So something like:
Try to do a graceful stop with timeout 30 seconds with `ContainerStop`. After that, the same code that exists today with the `ContainerKill` would execute which would forcefully kill the container.
Another reason for this is that after stopping the containers, the VM is anyways destroyed so we can be more lenient.

engine/docker/docker.go
vistaarjuneja: IMO adding a flag in the container spec which gets propagated from the IACM step would be helpful here (this would involve a change in the runner/java code as well). The APIs and methods could remain the same, but if `force` is set to `false` for a container in the Destroy config, we would first try to do a `ContainerStop` with a timeout as opposed to a `ContainerKill`.
This would give some soak time for the change in IACM and not affect the rest of the flows that exist today. Let me know your thoughts.

vistaarjuneja, line 160: Could we decrease the timeout here? I'm guessing the plugin has to make a network call so maybe something like 5-10 seconds would be better?

nassergonzalez: So, I like your idea but the problem that I see is that, the destroy step only have a parameter which is the pipelineConfig object. This seems to be a configuration object that is propagated from the runner to the lite-engine. The `ContainerRemoveOptions` seems to be a struct to control the removing of the containers, not the way to stop them.

With that said, i am experimenting now with something where, the InitialiseStep, which is the step  that contains the values for the `spec.PipelineConfig` would loop through the steps and if any of the steps is an apply or a destroy, then we set an envvar called `HARNESS_SOFT_STOP`. Then I could use that envvar which should be present in the destroy method to then choose if i want to perform a stop, or a kill. Then I was thinking to abstract the stop to another function so basically the code will be

if envar {
use stop method and don't finish until all the containers are stopped
} else {
kill as it was before
}

remove containers <- at this point if we executed the stop method all the containers should have been stopped.

By doing it with envvars i can keep the API methods the same although it feels a bit hacky to me.

If you know another way to do it I am happy to have a quick call and we can discuss it

engine/docker/docker.go
```diff
@@ -525,3 +537,33 @@ func (e *Docker) setProxyInDockerDaemon(ctx context.Context, pipelineConfig *spe
...
 	}
 }
+
+// SoftStop stops the container giving them a 30 seconds grace period. The signal sent by ContainerStop is SIGTERM.
+// After the grace period, the container is killed with SIGKILL.
+// After all the containers are stopped, they are removed only when the status is not "running" or "removing".
+func (e *Docker) SoftStop(ctx context.Context, name string) {
+	timeout := 30 * time.Second
+	if err := e.client.ContainerStop(ctx, name, &timeout); err != nil {
+		logrus.WithField("container", name).WithField("error", err).Warnln("failed to stop the container")
```
allofthesepeople, line 547: if we haven't been able to stop the contain we should probably return as there's no point in continuing with the rest of the procedure.

engine/docker/docker.go
```diff
@@ -525,3 +537,33 @@ func (e *Docker) setProxyInDockerDaemon(ctx context.Context, pipelineConfig *spe
...
+	}
+
+	// Before removing the container we want to be sure that it's in a healthy state to be removed.
+
+	containerStatus, err := e.client.ContainerInspect(ctx, name)
```
allofthesepeople, line 552: think the overall block could be simplified and a little more readable with with guards & everything in the loop:

```suggestion
for {
	time.Sleep(1 * time.Second)
	containerStatus, err := e.client.ContainerInspect(ctx, name)
	if err != nil {
		logrus.WithField("container", name).WithField("error", err).Warnln("failed to retrieve container stats")
		continue
	}
	if containerStatus.State.Status == "removing" || containerStatus.State.Status == "running" {
		continue
	}
	// everything has stopped
	break
}
```

allofthesepeople (commented):
a few comments

engine/docker/docker.go
```diff
@@ -61,7 +61,12 @@ type Docker struct {
 	client     client.APIClient
 	hidePull   bool
 	mu         sync.Mutex
-	containers []string
+	containers []Containers
+}
+
+type Containers struct {
```
vistaarjuneja, line 67: Nit: should be `Container`

engine/docker/docker.go
```diff
@@ -61,7 +61,12 @@ type Docker struct {
...
+	id       string
```
vistaarjuneja, line 68: Let's make id and softStop public - `ID` and `SoftStop`

engine/docker/docker.go
```diff
@@ -158,16 +163,20 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
 	e.mu.Unlock()
 
 	// stop all containers
-	for _, ctrName := range containers {
-		if err := e.client.ContainerKill(ctx, ctrName, "9"); err != nil {
-			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to kill container")
+	for _, ctiner := range containers {
```
vistaarjuneja, line 166: Nit: `ctr` might be better name than `ctiner`?

engine/docker/docker.go
```diff
@@ -525,3 +537,33 @@ func (e *Docker) setProxyInDockerDaemon(ctx context.Context, pipelineConfig *spe
 		}
 	}
 }
+
+// SoftStop stops the container giving them a 30 seconds grace period. The signal sent by ContainerStop is SIGTERM.
+// After the grace period, the container is killed with SIGKILL.
+// After all the containers are stopped, they are removed only when the status is not "running" or "removing".
+func (e *Docker) SoftStop(ctx context.Context, name string) {
```
vistaarjuneja, line 544: Can we make this method small case `softStop` as this should be called through the destroy

engine/docker/docker.go
```diff
@@ -525,3 +537,33 @@ func (e *Docker) setProxyInDockerDaemon(ctx context.Context, pipelineConfig *spe
...
+	timeout := 30 * time.Second
```
vistaarjuneja, line 545: Can we move this value to the top of the file in the constants?

engine/docker/docker.go
```diff
@@ -525,3 +537,33 @@ func (e *Docker) setProxyInDockerDaemon(ctx context.Context, pipelineConfig *spe
...
+
+	containerStatus, err := e.client.ContainerInspect(ctx, name)
+
+	if err != nil {
+		logrus.WithField("container", name).WithField("error", err).Warnln("failed to retrieve container stats")
+	}
+
+	for {
+		if err != nil && containerStatus.State.Status == "removing" || containerStatus.State.Status == "running" {
+			time.Sleep(1 * time.Second)
```
vistaarjuneja, line 560: We should move all times into constants at the top of the file

engine/docker/docker.go
```diff
@@ -525,3 +537,33 @@ func (e *Docker) setProxyInDockerDaemon(ctx context.Context, pipelineConfig *spe
...
+
+	// Before removing the container we want to be sure that it's in a healthy state to be removed.
+
+	containerStatus, err := e.client.ContainerInspect(ctx, name)
+
+	if err != nil {
+		logrus.WithField("container", name).WithField("error", err).Warnln("failed to retrieve container stats")
+	}
+
+	for {
```
vistaarjuneja, line 558: Just curious - is there anything preventing us from going in an infinite loop here in case `ContainerInspect` returns an error? Even though it should not happen, there should be some max retries.
Couldn't we force kill in case the `ContainerStop` happens successfully (would it exit out only after 30 seconds?). IMO if the container has been given 30 seconds to terminate gracefully, that should be enough time and we could terminate forcefully afterwards

engine/docker/docker.go
```diff
@@ -525,3 +537,33 @@ func (e *Docker) setProxyInDockerDaemon(ctx context.Context, pipelineConfig *spe
...
+		if err != nil && containerStatus.State.Status == "removing" || containerStatus.State.Status == "running" {
+			time.Sleep(1 * time.Second)
+			containerStatus, err = e.client.ContainerInspect(ctx, name)
```
vistaarjuneja, line 561: Can we move the `ContainerInspect` call only once inside the for loop and remove it from outside the for loop

engine/docker/docker.go
nassergonzalez: This will run for all containers, the ones that are already stopped because they were finalized too. Which means that they will return with err, which is why we don't return. We want to delete those containers from the docker daemon. 
Also for other containers that we can't stop, removing them will send a kill -9 ensuring that they are cleaned

engine/docker/docker.go
nassergonzalez: We can't pass the const address to the method so I can't put it in the const block

engine/docker/docker.go
nassergonzalez: Yeah makes sense to add something to be sure that even in the unlikely scenario of the 30seconds passing if we are still trying to do stuff, just break from the loop and continue

engine/docker/docker.go
```diff
@@ -525,3 +539,32 @@ func (e *Docker) setProxyInDockerDaemon(ctx context.Context, pipelineConfig *spe
...
+	}
+
+	// Before removing the container we want to be sure that it's in a healthy state to be removed.
+	now := time.Now()
+	for {
+		if time.Since(now) > timeout {
+			break
+		}
+		time.Sleep(1 * time.Second)
+		containerStatus, err := e.client.ContainerInspect(ctx, name)
```
vistaarjuneja, line 559: @nassergonzalez one question - doesn't the `timeout` you pass in `ContainerStop` take care of this logic? And do we want to try the stop only once?

engine/docker/docker.go
```diff
@@ -158,16 +165,20 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
+				logrus.WithField("container", ctr.ID).WithField("error", err).Warnln("failed to kill container")
+			}
 		}
 	}
 
 	// cleanup all containers
-	for _, ctrName := range containers {
-		if err := e.client.ContainerRemove(ctx, ctrName, removeOpts); err != nil {
-			logrus.WithField("container", ctrName).WithField("error", err).Warnln("failed to remove container")
+	for _, ctiner := range containers {
```
vistaarjuneja, line 179: Nit: `ctr` might be better than `ctiner`

engine/docker/docker.go
```diff
@@ -158,16 +165,20 @@ func (e *Docker) Destroy(ctx context.Context, pipelineConfig *spec.PipelineConfi
...
+		if err := e.client.ContainerRemove(ctx, ctiner.ID, removeOpts); err != nil {
+			logrus.WithField("container", ctiner).WithField("error", err).Warnln("failed to remove container")
```
vistaarjuneja, line 181: logrus.WithField("container", ctr.**ID**)

engine/docker/docker.go
nassergonzalez: No, so the timeout that we pass to container stop uses that to stop the container and if in 30 secs it was not kill, to kill it, but the call does not block so it returns immediately. I thought the same as you when i started to do my investigation and later on discovered that that was not the case.

And yes, i think that trying to stop once is enough. That will send the stop and the docker engine will kill it eventually. Checking the container inspect is to be sure that we do not try to remove the container while it's been removing or running state. If we do not have that, what happens is:
Docker sends the stop signal
Before the plugin can handle the stop signal
Docker removes the container

vistaarjuneja: got it

vistaarjuneja (approved):

PR #233, opened 2024-01-28
[CI-11171]: Parse and Upload Savings

rutvijmehta-harness:


ti/config/config.go
```diff
@@ -22,6 +23,7 @@ func New(endpoint, token, accountID, orgID, projectID, pipelineID, buildID, stag
 		commitBranch: commitBranch,
 		dataDir:      dataDir,
 		ignoreInstr:  false,
+		parseSavings: parseSavings,
```
anurag-harness, line 26: Can we pass a ff map instead of each ff as a separate argument? That way we wont have to keep updating the argument list when we add or deprecate ffs. What do the others think about this? @hemanthmantri @jamie-harness

ti/savings/savings.go
```diff
@@ -0,0 +1,31 @@
...
+	tiCfg "github.com/harness/lite-engine/ti/config"
+	"github.com/harness/lite-engine/ti/savings/cache"
+	"github.com/harness/ti-client/types"
+	"github.com/sirupsen/logrus"
+)
+
+func ParseAndUploadSavings(ctx context.Context, workspace string, log *logrus.Logger, stepID string, tiConfig *tiCfg.Cfg) {
+	// Cache Savings
+	start := time.Now()
+	state, timeTaken, err := cache.ParseCacheSavings(workspace, log)
```
anurag-harness, line 17: For TI timesavings we use the time taken by the step, but over here we are getting build time from the gradle profile report. 
What approach should be used for build cache? For build cache will the report time be same as the time taken for the step?

ti/savings/types.go
```diff
@@ -0,0 +1,17 @@
...
+
+import "github.com/harness/ti-client/types"
+
+type CacheType int32
+
+const (
+	GRADLE CacheType = 1
+)
+
+type Overview struct {
```
anurag-harness, line 11: Where is this being used?

ti/savings/cache/gradle/parser.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/mattn/go-zglob"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/net/html"
+)
+
+func ParseSavings(workspace string, log *logrus.Logger, logErr bool) (types.IntelligenceExecutionState, int, error) {
+	cacheState := types.FULL_RUN
+	totalBuildTime := 0
+
+	path := fmt.Sprintf("%s/build/reports/profile/*.html", workspace)
```
anurag-harness, line 21: Should we take the path out as a const?

ti/savings/cache/gradle/parser.go
```diff
@@ -0,0 +1,249 @@
...
+	files, err := zglob.Glob(path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return cacheState, totalBuildTime, fmt.Errorf("no profiles present")
+		}
+		if logErr {
```
anurag-harness, line 27: This will only be hit if its not a file exists error and we want to log. Why do we want to selectively log? When will it be used?

ti/savings/cache/gradle/parser.go
```diff
@@ -0,0 +1,249 @@
...
+
+	return totalBuildTime, nil
+}
+
+func extractValueFromTable(n *html.Node, description string) (value string) {
+	if n == nil {
+		return value
+	}
+	var traverse func(*html.Node)
+	traverse = func(node *html.Node) {
```
anurag-harness, line 93: Can this function be written better and with some comments?

anurag-harness (commented):
Havent reviewed the profile parsing logic, will take a closer look in a while

ti/config/config.go
rutvijmehta-harness: IMO step interface should be independent of any feature flags. Adding FF map here would just make it error prone

ti/savings/cache/gradle/parser.go
rutvijmehta-harness: Done

ti/savings/cache/gradle/parser.go
rutvijmehta-harness: Removed `logErr` boolean. If profile is not present, we'll try to look for other build caches. If present and there's some issue with the path, we need to log it.

ti/savings/cache/gradle/parser.go
rutvijmehta-harness: Updated with comment. Lmk what could be written better

ti/savings/types.go
rutvijmehta-harness: It'll be used in the feature. Check CI-10763

ti/savings/savings.go
rutvijmehta-harness: Yes that's the approach we want to take. Gradle build time is what matters for its time savings.

jamie-harness (approved):

PR #234, opened 2024-02-02
Add time savings for TI

rutvijmehta-harness:


ti/instrumentation/instrumentation.go
```diff
@@ -84,6 +84,14 @@ func getTestSelection(ctx context.Context, runner TestRunner, config *api.RunTes
 	} else {
 		log.Infoln(fmt.Sprintf("Running tests selected by Test Intelligence: %s", selection.Tests))
 	}
+
+	if tiConfig.GetParseSavings() {
+		if config.RunOnlySelectedTests {
+			tiConfig.WriteSavingsState(stepID, ti.TI, ti.OPTIMIZED)
+		} else {
+			tiConfig.WriteSavingsState(stepID, ti.TI, ti.FULL_RUN)
```
hemanthmantri, line 92: This is the case where you haven't set any optimization features (disabled TI) so it should be ti.DISABLED. FULL_RUN is when config.RunOnlySelectedTests is set (user enabled TI) but we ran all tests (selected == total). Let me know if I got this wrong.

ti/savings/savings.go
```diff
@@ -11,21 +11,36 @@ import (
...
+		tiErr := tiConfig.GetClient().WriteSavings(ctx, stepID, types.BUILD_CACHE, cacheState, int64(timeTaken))
 		if tiErr == nil {
 			log.Infof("Successfully uploaded savings for feature %s in %0.2f seconds",
 				types.BUILD_CACHE, time.Since(tiStart).Seconds())
 		}
 	}
+
+	// TI Savings
+	if tiState, err := tiConfig.GetSavingsState(stepID, types.TI); err == nil {
+		log.Infof("Successfully parsed Test Intelligence savings with state %s and time %dms",
```
hemanthmantri, line 33: s/parsed/computed

ti/savings/savings.go
```diff
@@ -11,21 +11,36 @@ import (
...
+func ParseAndUploadSavings(ctx context.Context, workspace string, log *logrus.Logger, stepID string,
+	cmdTimeTaken int64, tiConfig *tiCfg.Cfg) {
 	// Cache Savings
 	start := time.Now()
-	state, timeTaken, err := cache.ParseCacheSavings(workspace, log)
+	cacheState, timeTaken, err := cache.ParseCacheSavings(workspace, log)
 	if err == nil {
-		log.Infof("Successfully parsed savings with state %s and time %sms in %0.2f seconds",
-			state, strconv.Itoa(timeTaken), time.Since(start).Seconds())
+		log.Infof("Successfully parsed Build Cache savings with state %s and time %sms in %0.2f seconds",
```
hemanthmantri, line 20: s/parsed/computed. Let's not use `savings` here when we are printing the total time taken (not savings). Consider: `Computed build cache execution details. State: , Time taken:`

ti/savings/savings.go
hemanthmantri: Consider: `Computed test intelligence execution details. State: , Time taken:`

hemanthmantri (changes requested):

ti/instrumentation/instrumentation.go
rutvijmehta-harness: We would only hit this code if user asked for TI. Ref: https://github.com/harness/lite-engine/blob/main/ti/instrumentation/instrumentation.go#L219

`config.RunOnlySelectedTests` would be `false` here in case we get an error with TI. You can check the above lines.

rutvijmehta-harness: Added comments to clear the confusion

ti/savings/savings.go
rutvijmehta-harness: Done

ti/savings/savings.go
rutvijmehta-harness: Done

ti/instrumentation/instrumentation.go
hemanthmantri: Ah okay. The config.RunOnlySelectedTests is confusing.

hemanthmantri (approved):

jamie-harness (approved):

PR #235, opened 2024-02-03
Add optimization state to step response

rutvijmehta-harness:
Changes:
1. Calculate step optimization state from all feature states
2. Return step optimization state as part of step response

### DIFF
--- a/ti/savings/savings.go
+++ b/ti/savings/savings.go
@@ -31,7 +31,7 @@ func ParseAndUploadSavings(ctx context.Context, workspace string, log *logrus.Lo
 	}
 
 	// TI Savings
-	if tiState, err := tiConfig.GetSavingsState(stepID, types.TI); err == nil {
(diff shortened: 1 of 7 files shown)


ti/savings/savings.go
```diff
@@ -43,4 +46,22 @@ func ParseAndUploadSavings(ctx context.Context, workspace string, log *logrus.Lo
 
 	// DLC Savings (Placeholder)
 	// Cache Intel savings (Placeholder)
+	return getStepState(states)
```
devkimittal, line 49: Don't understand it, Why is states an array/list?

rutvijmehta-harness: There are 4 features: Build Cache, TI, DLC and Cache Intel. Each feature will give us a state and `getStepState` helps us get the final state of the step from all the feature states.

devkimittal (approved):

PR #238, opened 2024-02-14
Ignore module if no test rules are found

rutvijmehta-harness:


Aishwarya-Lad (approved):
Looks good. Tested it on VM, works fine too.

PR #243, opened 2024-02-21
feat: [CI-9572]: Added support of new TI step for VM

ShobhitSingh11:


pipeline/runtime/filterfileutil.go
```diff
@@ -0,0 +1,51 @@
...
+	filterFile := fmt.Sprintf("%s/filter", filePath)
+	f, err := fs.Create(filterFile)
+	if err != nil {
+		log.Println(fmt.Sprintf("could not create file %s", filterFile), err)
+		return err
+	}
+
+	var data string
+	for i := range testResp {
+		test := testResp[i]
```
smjt-h, line 34: What if the response is nil, will there be empty or no filter file. How does it effect the behaviour of test selection

ShobhitSingh11: If there is no filter file we are simply returning around Line no 20. If filterfile is present but null then empty data will be written. Rest Java agent will take care on the basis of contents of filter file

pipeline/runtime/runtest.go
```diff
@@ -112,7 +116,7 @@ func executeRunTestStep(ctx context.Context, engine *engine.Engine, r *api.Start
 // collectRunTestData collects callgraph and test reports after executing the step
 func collectRunTestData(ctx context.Context, log *logrus.Logger, r *api.StartStepRequest, start time.Time, stepName string, tiConfig *tiCfg.Cfg) error {
 	cgStart := time.Now()
-	cgErr := collectCgFn(ctx, stepName, time.Since(start).Milliseconds(), log, cgStart, tiConfig)
+	cgErr := collectCgFn(ctx, stepName, time.Since(start).Milliseconds(), log, cgStart, tiConfig, cgDir)
```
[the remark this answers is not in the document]
ShobhitSingh11: Had to include one parameter in upload callgraph method (In upload.go) this method is being used in both runTest and runTestV2 so had to update

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+			log.Errorln("Unable to get changed files list for PR. Running all the tests.", "error", err)
+			RunOnlySelectedTests = false
+			return selection, false
+		}
+	}
+
+	filesWithpkg := java.ReadPkgs(log, fs, workspace, files)
+	selection, err = utils.SelectTests(ctx, workspace, filesWithpkg, RunOnlySelectedTests, stepID, fs, tiConfig)
+	if err != nil {
+		log.WithError(err).Errorln("There was some issue in trying to figure out tests to run. Running all the tests")
```
hemanthmantri, line 187: `An unexpected error occurred during test selection. Running all tests.`

hemanthmantri: How do we debug the failures? Should we have a env flag such as `TI_DEBUG` for printing debug logs (including error codes/strings)?

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+		return "", err
+	}
+	iniFile := fmt.Sprintf("%s/config.ini", iniFileDir)
+	data := fmt.Sprintf(`outDir: %s
+	logLevel: 0 
+	logConsole: false
+	writeTo: JSON
+	packageInference: true`, dir)
+
+	log.Infof("Attempting to write to %s with config:\n%s", iniFile, data)
```
hemanthmantri, line 221: `Writing to ...`

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+		log.WithError(err).Errorln(fmt.Sprintf("could not write %s to config file %s", data, iniFile))
+		return "", err
+	}
+
+	return iniFile, nil //path of config.ini file
+}
+
+func getPreCmd(tmpFilePath string, fs filesystem.FileSystem, log *logrus.Logger) (string, error) {
+	var preCmd string
+	iniFilePath, err := createJavaConfigFile(tmpFilePath, fs, log)
```
hemanthmantri, line 240: Nit: `configFilePath`

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+	err = downloadJavaAgent(ctx, tmpFilePath, fs, log)
+	if err != nil {
+		return nil, nil, nil, nil, nil, "", fmt.Errorf("failed to download Java agent")
+	}
+
+	commands := fmt.Sprintf("%s\n%s", preCmd, r.RunTestsV2.Command[0])
+	step := toStep(r)
+	step.Command = []string{commands}
+	step.Entrypoint = r.RunTestsV2.Entrypoint
+	setTiEnvVariables(step, tiConfig)
```
hemanthmantri, line 59: Can you not do the export of JAVA_TOOL_OPTIONS here instead of returning it from `getPreCmd` and sending explicitly to `step.Command` ?

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+
+	err = createFilterFile(ctx, fs, step.ID, r.WorkingDir, log, tiConfig, tmpFilePath)
+	if err != nil {
+		return nil, nil, nil, nil, nil, "", fmt.Errorf("error while creating filter file %s", err)
```
hemanthmantri, line 63: I think we should call it `selected_tests` file to make it readable.

pipeline/runtime/filterfileutil.go
```diff
@@ -0,0 +1,51 @@
...
+		if test.Pkg != "" && test.Class != "" && test.Method != "" {
+			data += fmt.Sprintf("%s.%s,%s\n", test.Pkg, test.Class, test.Method)
+		} else if test.Pkg != "" && test.Class != "" {
+			data += fmt.Sprintf("%s.%s\n", test.Pkg, test.Class)
+		} else if test.Pkg != "" {
+			data += fmt.Sprintf("%s\n", test.Pkg)
```
hemanthmantri, line 40: For Python/Ruby you may have empty package but non empty class (which is a file). cc: @jamie-harness

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+	crErr := report.ParseAndUploadTests(ctx, r.TestReport, r.WorkingDir, stepName, log, reportStart, tiConfig, r.Envs)
+	if crErr != nil {
+		log.WithField("error", crErr).Errorln(fmt.Sprintf("Failed to upload report. Time taken: %s", time.Since(reportStart)))
+	}
+	return cgErr
+}
+
+// Second parameter in return type (bool) is will be used to decide whether the filter file should be created or not.
+// In case of running all the cases no filter file should be created.
+func getTestsSelection(ctx context.Context, fs filesystem.FileSystem, stepID, workspace string, log *logrus.Logger,
```
rutvijmehta-harness, line 139: Can we reuse this method from the existing runtest step?

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+		log.Infoln("Test Intelligence determined to run all the tests")
+		RunOnlySelectedTests = false // TI selected all the tests to be run
+	} else {
+		log.Infoln(fmt.Sprintf("Running tests selected by Test Intelligence: %s", selection.Tests))
+	}
+
+	return selection, true
+}
+
+func createJavaConfigFile(tmpDir string, fs filesystem.FileSystem, log *logrus.Logger) (string, error) {
```
rutvijmehta-harness, line 199: Can we reuse this method from the existing runtest step?

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+		log.WithError(err).Errorln(fmt.Sprintf("failed to write in .bazelrc file"))
+		return "", err
+	}
+	javaAgentPath := fmt.Sprintf("%s%s%s", tmpFilePath, javaNewAgentPath, javaNewAgentJar)
+	agentArg := fmt.Sprintf(javaNewAgentArg, javaAgentPath, iniFilePath)
+	preCmd = fmt.Sprintf("export JAVA_TOOL_OPTIONS=%s", agentArg)
+	return preCmd, nil
+}
+
+func downloadJavaAgent(ctx context.Context, path string, fs filesystem.FileSystem, log *logrus.Logger) error {
```
rutvijmehta-harness, line 257: Can we reuse this method from the existing runtest step?

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+		log.WithError(err).Errorln(fmt.Sprintf("could not download java agent"))
+		return err
+	}
+	return nil
+}
+
+func createFilterFile(ctx context.Context, fs filesystem.FileSystem, stepID, workspace string, log *logrus.Logger,
+	tiConfig *tiCfg.Cfg, path string) error {
+
+	isManualExecution := utils.IsManualExecution(tiConfig)
```
rutvijmehta-harness, line 272: @anurag-harness how do we know if we do not want to attach instrumentation? We don't have `runOnlySelectedTests` as task input.

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	err = populateItemInFilterFile(resp, dir, fs, isFilterFilePresent)
+
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func writetoBazelrcFile(iniFilePath string, log *logrus.Logger, fs filesystem.FileSystem, tmpFilePath string) error {
```
rutvijmehta-harness, line 288: Should we look for bazelrc file in the test command? Users can give another file as an input. You can refer to PR checks pipeline.

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+		log.WithError(err).Errorln("There was some issue in trying to figure out tests to run. Running all the tests")
+		RunOnlySelectedTests = false
+	} else if selection.SelectAll {
+		log.Infoln("Test Intelligence determined to run all the tests")
+		RunOnlySelectedTests = false // TI selected all the tests to be run
+	} else {
+		log.Infoln(fmt.Sprintf("Running tests selected by Test Intelligence: %s", selection.Tests))
+	}
+
+	return selection, true
```
rutvijmehta-harness, line 196: Should you not be returning `RunOnlySelectedTests` instead of hardcoded values everywhere?

rutvijmehta-harness (commented):
Initial review

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+// Second parameter in return type (bool) is will be used to decide whether the filter file should be created or not.
+// In case of running all the cases no filter file should be created.
+func getTestsSelection(ctx context.Context, fs filesystem.FileSystem, stepID, workspace string, log *logrus.Logger,
+	isManual bool, tiConfig *tiCfg.Cfg) (types.SelectTestsResp, bool) {
+
+	selection := types.SelectTestsResp{}
+	var RunOnlySelectedTests = true // Enabled true by default
+
+	if isManual {
+		log.Infoln("Manual execution has been detected. Running all the tests")
```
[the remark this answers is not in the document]
anurag-harness: It will never hit the isManual case anymore

pipeline/runtime/runtestsV2.go
anurag-harness: Yeah we are always attaching instrumentation

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,330 @@
...
+	} else {
+		files, err = utils.GetChangedFilesPR(ctx, workspace, log)
+		if err != nil || len(files) == 0 {
+			log.Errorln("Unable to get changed files list for PR. Running all the tests.", "error", err)
+			RunOnlySelectedTests = false
+			return selection, false
+		}
+	}
+
+	filesWithpkg := java.ReadPkgs(log, fs, workspace, files)
```
anurag-harness, line 184: You can infer the file language fro the fil extension and if it is Java then call ReadPkgs for that file

pipeline/runtime/runtest.go
```diff
@@ -24,6 +24,10 @@ import (
 	easyFormatter "github.com/t-tomalak/logrus-easy-formatter"
 )
 
+const (
+	cgDir = "%s/ti/callgraph/" // path where callgraph files will be generated
```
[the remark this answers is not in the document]
ShobhitSingh11: Looking into that

pipeline/runtime/runtestsV2.go
ShobhitSingh11: tmpFilePath is being provided and tiConfig. This is /tmp/engine

pipeline/runtime/runtestsV2.go
ShobhitSingh11: Since for preCommand we needed iniFilePath being coming from javaAgentConfigFile and also same inifile is being used to write to .bazelrc file to support bazel execution. That's why we have enclosed in a method where config file is being created and env variable is being set

pipeline/runtime/runtestsV2.go
ShobhitSingh11: We could have but here some things have changed not eveything which is being used for runtest are used for runtestV2 because runner interface is being used for some calls there and parameters like "language", "build tool" etc are required in order to create runner. So in getTestsSelection some methods are being called by runner and also for **checkForBazelOptimization** it requires those params like "language", "build tool" . Even though many things are same there are still some changes. That's why thought of creating this method for V2

pipeline/runtime/runtestsV2.go
ShobhitSingh11: Same reason as above, runner dependency and some config from RunTestsConfig which is not being provided by RuntestsV2

pipeline/runtime/runtestsV2.go
ShobhitSingh11: This method is being written because the path for downloading agent is different from runtest step, we can see that 
ultimately it is calling DownloadFile from utils. As for now we are using one source only

pipeline/runtime/runtestsV2.go
ShobhitSingh11: In order to support for bazel, env variable had to be exported in .bazelrc file else overridden of env var takes place which is being used to inject java agent. We can have this support in .bazelrc in home directory level. @smjt-h had tested this case.

pipeline/runtime/runtestsV2.go
ShobhitSingh11: Have removed RunOnlySelectedTests. What is the use of that boolean return has been mentioned on the top of this method

pipeline/runtime/runtestsV2.go
ShobhitSingh11: Looking into that

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,332 @@
+// Copyright 2024 Drone.IO Inc. All rights reserved.
+// Use of this source code is governed by the Polyform License
```
anurag-harness, line 2: We should add a boolean to disable all instrumentation

pipeline/runtime/runtestsV2.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return nil, nil, nil, nil, nil, string(optimizationState), fmt.Errorf("failed to download Java agent")
+	}
+
+	commands := fmt.Sprintf("%s\n%s", preCmd, r.RunTestsV2.Command[0])
+	step := toStep(r)
+	step.Command = []string{commands}
+	step.Entrypoint = r.RunTestsV2.Entrypoint
+	setTiEnvVariables(step, tiConfig)
+	err = createSelectedTestFile(ctx, fs, step.ID, r.WorkingDir, log, tiConfig, tmpFilePath, r.Envs, &r.RunTestsV2, filterfilePath)
```
anurag-harness, line 63: [Important] Change to step.Name instead of step.ID

PR #245, opened 2024-03-08
[feat]: [CI-11159]: Support Ruby for New RunStep

anurag-harness:


pipeline/runtime/runtestsV2.go
```diff
@@ -264,25 +276,51 @@ func getPreCmd(tmpFilePath string, fs filesystem.FileSystem, log *logrus.Logger,
 	javaAgentPath := fmt.Sprintf("%s%s%s", tmpFilePath, javaAgentV2Path, javaAgentV2Jar)
 	agentArg := fmt.Sprintf(javaAgentV2Arg, javaAgentPath, iniFilePath)
 	preCmd = fmt.Sprintf("export JAVA_TOOL_OPTIONS=%s export BAZEL_SYSTEM_BAZELRC_PATH=%s", agentArg, bazelfilepath)
+
+	// Ruby
+	envs["TI"] = "1"
+	envs["TI_V2"] = "1"
```
jamie-harness, line 282: Do we need this? The original 'TI` is for the script we automatically adding to spec_helper.rb

anurag-harness: We use the original TI in the agent as well

PR #258, opened 2024-03-31
[feat]: [CI-11493]: fix bazel dir

smjt-h:


pipeline/runtime/runtestsV2.go
```diff
@@ -270,14 +269,14 @@ func getPreCmd(workspace, tmpFilePath string, fs filesystem.FileSystem, log *log
 		return "", "", err
 	}
 
-	bazelfilepath, err := writetoBazelrcFile(log, fs, tmpFilePath, splitIdx)
+	_, err = writetoBazelrcFile(log, fs, tmpFilePath, splitIdx)
```
ShobhitSingh11, line 272: We can change the method signature of returning type error only if we are not using string return type anywhere.

ShobhitSingh11 (changes requested):

ShobhitSingh11 (approved):

pipeline/runtime/runtestsV2.go
```diff
@@ -348,48 +347,46 @@ func createSelectedTestFile(ctx context.Context, fs filesystem.FileSystem, stepI
...
 		if err != nil {
 			log.WithError(err).Errorln(fmt.Sprintf("could not open the file in dir %s", bazelrcFilePath))
-			return "", err
+			return err
 		}
 		defer file.Close()
 
 		log.Printf(fmt.Sprintf("attempting to write %s to %s", data, bazelrcFilePath))
-		_, err = file.WriteString(data)
+		_, err = file.WriteString("\n" + data)
```
ShobhitSingh11, line 383: LGTM,  we need to make sure "\n" + data is working

PR #260, opened 2024-04-01
[CI-11819]: Add exclude filter for vendor

jamie-harness:


ti/instrumentation/utils.go
```diff
@@ -451,6 +454,18 @@ func filterTestsAfterSelection(selection ti.SelectTestsResp, testGlobs string) t
 	return selection
 }
 
+func isExcluded(class string) bool {
+	if os.Getenv("TI_SKIP_EXCLUDE_VENDOR") == "true" {
```
anurag-harness, line 458: read the env map of the step to get the env variables

anurag-harness (changes requested):

anurag-harness:
This should be added in the flow for AutoDetect Tests for Parallelism as well

ti/instrumentation/utils.go
jamie-harness: done

anurag-harness (approved):

PR #264, opened 2024-04-07
[CI-11492]: Add autodetection splitting

smjt-h:
Added auto-test-detection for java, ruby and python for test splitting and parallelism.
In case of prallelism if its first run or if all cases are selected then tests will be auto detected and populated in filter file.

### DIFF
--- a/pipeline/runtime/common.go
+++ b/pipeline/runtime/common.go
@@ -197,26 +197,9 @@ func waitForFileWithTimeout(timeout time.Duration, tiConfig *tiCfg.Cfg) error {
 		if !tiConfig.IsLockedRuby() {
 			return nil
 		}
-
 		if time.Now().After(deadline) {
 			return fmt.Errorf("timeout waiting for agent download")
 		}
-
-		time.Sleep(time.Millisecond * 100)
-	}
-}
-
-func waitForFileWithTimeoutPy(timeout time.Duration, tiConfig *tiCfg.Cfg) error {
-	deadline := time.Now().Add(timeout)
-	for {
-		if !tiConfig.IsLockedPython() {
-			return nil
-		}
-
-		if time.Now().After(deadline) {
-			return fmt.Errorf("timeout waiting for agent download")
-		}
-
 		time.Sleep(time.Millisecond * 100)
 	}
 }
--- a/pipeline/runtime/common.go
+++ b/pipeline/runtime/common.go
@@ -194,12 +194,12 @@ func GetReplacer(
 func waitForFileWithTimeout(timeout time.Duration, tiConfig *tiCfg.Cfg) error {
 	deadline := time.Now().Add(timeout)
 	for {
+		time.Sleep(time.Second * 1)
 		if !tiConfig.IsZipLocked() {
 			return nil
 		}
 		if time.Now().After(deadline) {
 			return fmt.Errorf("timeout waiting for agent download")
 		}
-		time.Sleep(time.Millisecond * 100)
 	}
 }
--- a/pipeline/runtime/common.go
+++ b/pipeline/runtime/common.go
@@ -17,7 +17,6 @@ import (
 	v3 "github.com/harness/godotenv/v3"
 	"github.com/harness/lite-engine/api"
 	"github.com/harness/lite-engine/engine/spec"
-	"github.com/harness/lite-engine/internal/filesystem"
 	"github.com/harness/lite-engine/livelog"
 	"github.com/harness/lite-engine/logstream"
 	"github.com/harness/lite-engine/logstream/remote"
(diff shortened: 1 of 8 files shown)


pipeline/runtime/runtestsV2.go
```diff
@@ -203,20 +201,21 @@ func getTestsSelection(ctx context.Context, fs filesystem.FileSystem, stepID, wo
 	selection, err = instrumentation.SelectTests(ctx, workspace, filesWithpkg, runOnlySelectedTests, stepID, fs, tiConfig)
 	if err != nil {
 		log.WithError(err).Errorln("An unexpected error occurred during test selection. Running all tests.")
-		return selection, false
+		runOnlySelectedTests = false
```
ShobhitSingh11, line 204: We should make sure, normal test selection is working as we are removing return statement here.

pipeline/runtime/runtestsV2.go
```diff
@@ -203,20 +201,21 @@ func getTestsSelection(ctx context.Context, fs filesystem.FileSystem, stepID, wo
...
 	} else if selection.SelectAll {
 		log.Infoln("Test Intelligence determined to run all the tests")
-		return selection, false
+		runOnlySelectedTests = false
```
ShobhitSingh11, line 207: We should make sure, normal test selection is working as we are removing return statement here.

ti/instrumentation/instrumentation.go
```diff
@@ -132,23 +135,44 @@ func checkForBazelOptimization(ctx context.Context, workspace string, fs filesys
...
+			// receives error / no tests to run, all the other steps should have the same output
+			if splitIdx == 0 {
+				// Error while auto-detecting, run all tests for parallel step 0
+				runOnlySelectedTests = false
+				log.Errorln("Error in auto-detecting tests for splitting, running all tests")
+			} else {
+				// Error while auto-detecting, no tests for other parallel steps
+				selection.Tests = []ti.RunnableTest{}
+				runOnlySelectedTests = true
+				log.WithError(err).Errorln("Error in auto-detecting tests for splitting, running all tests in parallel step 0")
```
ShobhitSingh11, line 166: parallel step 0 ??

ShobhitSingh11 (changes requested):

ti/instrumentation/instrumentation.go
smjt-h: yes, in case of any errors during test selection , the 0th step should run everything, other steps will run no tests

pipeline/runtime/runtestsV2.go
smjt-h: yes tested with test selection as well

ti/instrumentation/instrumentation.go
```diff
@@ -132,23 +135,44 @@ func checkForBazelOptimization(ctx context.Context, workspace string, fs filesys
...
+			// AutoDetectTests output should be same across all the parallel steps. If one of the step
+			// receives error / no tests to run, all the other steps should have the same output
+			if splitIdx == 0 {
+				// Error while auto-detecting, run all tests for parallel step 0
+				runOnlySelectedTests = false
+				log.Errorln("Error in auto-detecting tests for splitting, running all tests")
+			} else {
+				// Error while auto-detecting, no tests for other parallel steps
+				selection.Tests = []ti.RunnableTest{}
+				runOnlySelectedTests = true
```
ShobhitSingh11, line 165: why we are doing "runOnlySelectedTests = true" here

ShobhitSingh11 (changes requested):

ti/instrumentation/instrumentation.go
smjt-h: so that, in case there is any error during test selection and its not 0th step during parallelism, it should not run anything, filter file will be populated with 0 cases.

pipeline/runtime/common.go
```diff
@@ -189,3 +191,19 @@ func GetReplacer(
 	wc := livelog.New(client, logKey, name, []logstream.Nudge{}, false)
 	return logstream.NewReplacer(wc, secrets)
 }
+
+func waitForFileWithTimeout(timeout time.Duration, filename string, fs filesystem.FileSystem) error {
```
anurag-harness, line 195: We dont need file for syncing in VM. We can use mutexes since there is one lite-engine context per stage. Check this usage and do something similar: https://github.com/harness/lite-engine/blob/ca535efeeb31348781da9d77dcc94130fe8b45c5/ti/config/config.go#L21

anurag-harness (changes requested):

pipeline/runtime/common.go
smjt-h: fixed

ti/config/config.go
```diff
@@ -127,3 +128,14 @@ func (c *Cfg) GetFeatureState(stepID string, feature types.SavingsFeature) (type
 	}
 	return types.DISABLED, ErrStateNotFound
 }
+
+func (c *Cfg) LockZipForRuby() {
```
anurag-harness, line 132: Access to this variable rubylocked should be done using a mutex since multiple steps will be reading and writing to it at the same time

smjt-h: Anyway I had to use an extra variable to check the status of lock, as in go there is no direct way to check the status of lock so added this directly,
I will add a mutex as well

ti/config/config.go
```diff
@@ -127,3 +130,16 @@ func (c *Cfg) GetFeatureState(stepID string, feature types.SavingsFeature) (type
 	}
 	return types.DISABLED, ErrStateNotFound
 }
+
+func (c *Cfg) LockZip() {
+	c.zipmu.Lock()
+	atomic.StoreInt32(&c.ziplocked, 1)
```
anurag-harness, line 136: If you are using atomic operations then you dont need then mutex

anurag-harness (approved):

ti/config/config.go
smjt-h: modified logic with atomic operations itself

PR #265, opened 2024-04-10
[CI-11998]: add intelligence mode

smjt-h:


pipeline/runtime/runtestsV2.go
```diff
@@ -49,39 +49,46 @@ func executeRunTestsV2Step(ctx context.Context, f RunFunc, r *api.StartStepReque
...
-		return nil, nil, nil, nil, nil, string(optimizationState), fmt.Errorf("error while creating filter file %s", err)
+		preCmd, filterfilePath, err := getPreCmd(r.WorkingDir, tmpFilePath, fs, log, r.Envs, agentPaths)
+		if err != nil || pythonArtifactDir == "" {
+			return nil, nil, nil, nil, nil, string(optimizationState), fmt.Errorf("failed to set config file or env variable to inject agent, %s", err)
+		}
+
+		commands := fmt.Sprintf("%s\n%s", preCmd, r.RunTestsV2.Command[0])
+		step.Command = []string{commands}
+
+		setTiEnvVariables(step, tiConfig)
```
ShobhitSingh11, line 82: Do we need to set TiEnvVariables only if Intelligence mode is on ?  I can see in runTest step or run step it is getting set irrespective of any condition.

ShobhitSingh11 (changes requested):

pipeline/runtime/runtestsV2.go
smjt-h: yes, this should be set always, moved this above

ShobhitSingh11 (approved):
LGTM

PR #278, opened 2024-05-03
fix:[CI-11792]: updated godotenv version

sahithibanda01:


pipeline/runtime/common.go
```diff
@@ -117,6 +117,9 @@ func fetchExportedVarsFromEnvFile(envFile string, out io.Writer, useCINewGodotEn
 
 	if useCINewGodotEnvVersion {
 		env, err = v3.Read(envFile)
+		if err != nil {
```
raghavharness, line 120: Add a comment above mentioning fallback to old godotenv version

harshjain12 (approved):

pipeline/runtime/common.go
harshjain12: @raghavharness sorry, I missed your comment

PR #279, opened 2024-05-06
[feat]: [CI-12466]: Initialize optimizationState to empty string

anurag-harness:
optimizationState should be initialized to an empty string and should only contain a non empty value if the FF to get optimization state is enabled. This will ensure backward compatibility of a vm runner with an older delegate

### DIFF
--- a/pipeline/runtime/run.go
+++ b/pipeline/runtime/run.go
@@ -32,7 +32,7 @@ func executeRunStep(ctx context.Context, f RunFunc, r *api.StartStepRequest, out
 	step.Entrypoint = r.Run.Entrypoint
 	setTiEnvVariables(step, tiConfig)
 
-	optimizationState := types.DISABLED
+	var optimizationState types.IntelligenceExecutionState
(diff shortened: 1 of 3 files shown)


rutvijmehta-harness (changes requested):
@anurag-harness why is this change required? The empty string will still be part of the JSON. Backward compatibility is taken care of by `ignoreUnknown` annotation.

If the delegate is very old, we don't guarantee backward compatibility anyway. Please check with delegate team the number of releases after which backward compatibility is not supported/guaranteed.

anurag-harness:
> @anurag-harness why is this change required? The empty string will still be part of the JSON. Backward compatibility is taken care of by `ignoreUnknown` annotation.
> 
> If the delegate is very old, we don't guarantee backward compatibility anyway. Please check with delegate team the number of releases after which backward compatibility is not supported/guaranteed.

@rutvijmehta-harness Due to the omitempty in the structure, the field will be omitted if it is an empty string

PR #280, opened 2024-05-09
[feat]: [CI-12471]: Send testglobs from lite-engine to ti-service

anurag-harness:
Currently test globs passed from the user in the runtest and test steps are not being sent to ti-service. TI service needs them to check if a changed file is a test file or no

Testing for vagrant ruby repo which does not use default globs

Before changes, if a test file is modified, it is not selected to run
Screenshot 2024-05-09 at 3 05 48 PM

With changes, modfying a test file we can see tests are selected (prints look different for debugging)
Screenshot 2024-05-09 at 2 58 26 PM

### DIFF
--- a/pipeline/runtime/runtestsV2.go
+++ b/pipeline/runtime/runtestsV2.go
@@ -511,7 +511,11 @@ func sanitizeTestGlobsV2(globStrings []string) []string {
 	for _, globString := range globStrings {
 		if globString != "" {
 			splitted := strings.Split(globString, ",")
-			result = append(result, splitted...)
+			for _, s := range splitted {
+				if s != "" {
+					result = append(result, s)
+				}
+			}
 		}
 	}
 	return result
--- a/pipeline/runtime/runtestsV2.go
+++ b/pipeline/runtime/runtestsV2.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/drone/runner-go/pipeline/runtime"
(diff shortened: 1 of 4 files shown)


jamie-harness (approved):

pipeline/runtime/runtestsV2.go
```diff
@@ -503,3 +505,18 @@ func collectTestReportsAndCg(ctx context.Context, log *logrus.Logger, r *api.Sta
 	}
 	return cgErr
 }
+
+func sanitizeTestGlobsV2(globStrings []string) []string {
+	var result = make([]string, 0)
+	for _, globString := range globStrings {
+		if globString != "" {
+			splitted := strings.Split(globString, ",")
```
hemanthmantri, line 513: splits

pipeline/runtime/runtestsV2.go
```diff
@@ -503,3 +505,18 @@ func collectTestReportsAndCg(ctx context.Context, log *logrus.Logger, r *api.Sta
 	}
 	return cgErr
 }
+
+func sanitizeTestGlobsV2(globStrings []string) []string {
```
hemanthmantri, line 509: Please add doc string. Explain params and return values.

pipeline/runtime/runtestsV2.go
hemanthmantri: There was a case where we had a comma in "glob" which would break this. Let's make sure we cover such cases in input validation - maybe it was a Ruby pattern and not a glob.

anurag-harness: Lets treat it as a separate task since we will have to change the sanitization function of the current RunTestStep as well

PR #282, opened 2024-05-15
[feat]: [CI-12562]: initial commit support for secrets

Sapa96:


pipeline/runtime/run.go
```diff
@@ -110,12 +117,48 @@ func executeRunStep(ctx context.Context, f RunFunc, r *api.StartStepRequest, out
...
+			for _, key := range r.OutputVars {
+				if _, ok := outputs[key]; ok {
+					output := &api.OutputV2{
+						Key:   key,
+						Value: outputs[key],
+						Type:  outputVariableTypeString,
+					}
+					outputsV2 = append(outputsV2, output)
+				}
+			}
```
raghavharness, line 132: Why do we need to add this ?
OutputVars is deprecated so we can skip this

pipeline/runtime/run.go
```diff
@@ -110,12 +117,48 @@ func executeRunStep(ctx context.Context, f RunFunc, r *api.StartStepRequest, out
...
+		} else {
```
raghavharness, line 133: Why this else cond is needed ?

pipeline/runtime/runtest.go
```diff
@@ -91,10 +91,14 @@ func executeRunTestStep(ctx context.Context, f RunFunc, r *api.StartStepRequest,
 	}
 	exportEnvs, _ := fetchExportedVarsFromEnvFile(exportEnvFile, out, useCINewGodotEnvVersion)
 	artifact, _ := fetchArtifactDataFromArtifactFile(artifactFile, out)
-	if len(r.Outputs) > 0 {
```
raghavharness, line 94: Why changes are needed in runtest ? run.go should be used for plugins right ?

raghavharness (changes requested):

pipeline/runtime/runtestsV2.go
```diff
@@ -106,6 +106,9 @@ func executeRunTestsV2Step(ctx context.Context, f RunFunc, r *api.StartStepReque
 	outputFile := fmt.Sprintf("%s/%s-output.env", pipeline.SharedVolPath, step.ID)
 	step.Envs["DRONE_OUTPUT"] = outputFile
 
+	outputSecretsFile := fmt.Sprintf("%s/%s-output-secrets.env", pipeline.SharedVolPath, step.ID)
```
raghavharness, line 109: Same for this

pipeline/runtime/run.go
Sapa96: pipelines with older yaml can still use this OutputVars fields right ? added this code for backward compatibility

pipeline/runtime/run.go
Sapa96: when variables are exported from plugin . both the conditions will not satisfy . so the flow will need to go to else case to read them 
```
if len(r.Outputs) > 0 
len(r.OutputVars) > 0 
```

pipeline/runtime/runtest.go
Sapa96: this is just a refactor of logic to match the format of run.go and runtest.go for easier readability .

pipeline/runtime/runtestsV2.go
Sapa96: observed the following support in the file . so added logic for secrets here also . can be removed if not required
`step.Envs["DRONE_OUTPUT"] = outputFile`

pipeline/runtime/run.go
raghavharness: I think you can remove this elseif cond altogether

pipeline/runtime/runtestsV2.go
raghavharness: I think let's do for run.go for now only ? If the requirement comes for runtest we can add that later ?

raghavharness (changes requested):

pipeline/runtime/run.go
```diff
@@ -66,6 +67,9 @@ func executeRunStep(ctx context.Context, f RunFunc, r *api.StartStepRequest, out
 		}
 	}
 
+	outputSecretsFile := fmt.Sprintf("%s/%s-output-secrets.env", pipeline.SharedVolPath, step.ID)
```
vistaarjuneja, line 70: Can you add the same condition as L52 and add a `SecretVarFile` variable. If it's set, we will use that directly otherwise generate it similarly to what you have. This will help integrate with runner in the future.

vistaarjuneja:
```
if r.SecretVarFile != "" {
     // set the env variable to r.SecretVarFile
}
```

pipeline/runtime/run.go
```diff
@@ -110,12 +117,48 @@ func executeRunStep(ctx context.Context, f RunFunc, r *api.StartStepRequest, out
...
+					Value: value,
+					Type:  outputVariableTypeString,
+				}
+				outputsV2 = append(outputsV2, output)
+			}
 		}
-		return exited, outputs, exportEnvs, artifact, nil, string(optimizationState), nil
+
+		//checking exported secrets from plugins if any
+		_, secretErr := os.Stat(outputSecretsFile)
```
vistaarjuneja, line 145: Let's log if `secretErr != nil`

pipeline/runtime/run.go
```diff
@@ -110,12 +117,48 @@ func executeRunStep(ctx context.Context, f RunFunc, r *api.StartStepRequest, out
...
+		if secretErr == nil {
+			secrets, err := fetchExportedVarsFromEnvFile(outputSecretsFile, out, useCINewGodotEnvVersion)
```
vistaarjuneja, line 147: If this errors, we should not iterate over the secrets right?
Let's follow golang convention of checking for error immediately afterwards.

pipeline/runtime/runtest.go
vistaarjuneja: I agree, we can skip adding in runtest.
The long term goal is for test intelligence to just be a plugin and to remove run tests special implementation from here.

pipeline/runtime/common.go
```diff
@@ -27,7 +27,9 @@ import (
 )
 
 const (
-	ciNewVersionGodotEnv = "CI_NEW_VERSION_GODOTENV"
+	ciNewVersionGodotEnv     = "CI_NEW_VERSION_GODOTENV"
+	outputVariableTypeString = "STRING"
```
vistaarjuneja, line 31: Can we make this an enum instead and define this in `api.go` so the contract is clear?

pipeline/runtime/run.go
Sapa96: @vistaarjuneja 
Why do we need to log it ? this will return err for all the existing plugins as they dont use secret file as of now . this check is to understand whether a secret is exported or not . 
if we log it , the error will be logged for all the existing plugins .

pipeline/runtime/common.go
Sapa96: done

pipeline/runtime/run.go
Sapa96: done

api/api.go
```diff
@@ -215,3 +219,10 @@ const (
 	Success CommandExecutionStatus = "SUCCESS"
 	Failure CommandExecutionStatus = "FAILURE"
 )
+
+type OutputType string
+
+const (
+	STRING OutputType = "STRING"
```
vistaarjuneja, line 226: Let's renamename these so it's more clear what it does
```
      OutputTypeString OutputType = "STRING"
      OutputTypeSecret OutputType = "SECRET"
```
so `api.OutputTypeString` is more clear

api/api.go
```diff
@@ -215,3 +219,10 @@ const (
 	Success CommandExecutionStatus = "SUCCESS"
 	Failure CommandExecutionStatus = "FAILURE"
 )
+
+type OutputType string
```
vistaarjuneja, line 223: change `Type` in `OutputV2` to be of type `OutputType`

pipeline/runtime/run.go
vistaarjuneja: Ack, let's change it to this in a single line:
```
if _, err := os.Stat(outputSecretsFile); err == nil {
  
}
```

vistaarjuneja (approved):

pipeline/runtime/run.go
```diff
@@ -99,8 +109,14 @@ func executeRunStep(ctx context.Context, f RunFunc, r *api.StartStepRequest, out
 	artifact, _ := fetchArtifactDataFromArtifactFile(artifactFile, out)
 	if exited != nil && exited.Exited && exited.ExitCode == 0 {
 		outputs, err := fetchExportedVarsFromEnvFile(outputFile, out, useCINewGodotEnvVersion) //nolint:govet
+		if err != nil {
+			log.WithError(err).Errorln("error encountered while fetching secrets from env File")
```
raghavharness, line 113: Can you fix the error msg, it's not for errors

raghavharness (changes requested):

PR #285, opened 2024-05-28
[feat]: [CI-12621]: Upload savings for DLC

anurag-harness:
We will parse the cache metrics file created by the buildx plugin if it exists and upload savings data to ti-service

### DIFF
--- a/ti/savings/dlc/dlc.go
+++ b/ti/savings/dlc/dlc.go
@@ -33,21 +33,18 @@ func GetFeatureState(cacheMetricsFile string, log *logrus.Logger) (types.Intelli
 
 	// Check if the file exists.
 	if _, err := os.Stat(cacheMetricsFile); os.IsNotExist(err) {
-		log.WithField("file", cacheMetricsFile).Error("Cache metrics file does not exist")
 		return state, err
 	}
 
 	// Read the JSON file containing the cache metrics.
 	data, err := os.ReadFile(cacheMetricsFile)
 	if err != nil {
-		log.WithField("file", cacheMetricsFile).Error("Failed to read cache metrics file")
 		return state, err
 	}
 
 	// Deserialize the JSON data into the CacheMetrics struct.
 	var metrics CacheMetrics
 	if err := json.Unmarshal(data, &metrics); err != nil {
-		log.WithField("file", cacheMetricsFile).Error("Failed to unmarshal cache metrics data")
 		return state, err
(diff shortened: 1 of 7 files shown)


ti/savings/savings.go
```diff
@@ -44,7 +46,22 @@ func ParseAndUploadSavings(ctx context.Context, workspace string, log *logrus.Lo
...
-	// DLC Savings (Placeholder)
+	// DLC Savings
+	if cacheMetricsFile, found := envs["PLUGIN_CACHE_METRICS_FILE"]; found {
+		if opts, ok := envs["PLUGIN_BUILDER_DRIVER_OPTS"]; ok && strings.Contains(opts, "harness/buildkit") {
+			dlcState, err := dlc.GetFeatureState(cacheMetricsFile, log)
+			if err == nil {
+				states = append(states, dlcState)
+				log.Infof("Computed docker layer caching execution details with state %s and time %dms", dlcState, cmdTimeTaken)
+				tiStart := time.Now()
+				tiErr := tiConfig.GetClient().WriteSavings(ctx, stepID, types.DLC, dlcState, cmdTimeTaken, types.SavingsRequest{})
```
rutvijmehta-harness, line 57: If you're already parsing the data, it doesn't hurt to send the savings details as well. We can add the server side changes later for DLC

ti/savings/dlc/dlc.go
```diff
@@ -0,0 +1,64 @@
...
+	data, err := os.ReadFile(cacheMetricsFile)
+	if err != nil {
+		log.WithField("file", cacheMetricsFile).Error("Failed to read cache metrics file")
+		return state, err
+	}
+
+	// Deserialize the JSON data into the CacheMetrics struct.
+	var metrics CacheMetrics
+	if err := json.Unmarshal(data, &metrics); err != nil {
+		log.WithField("file", cacheMetricsFile).Error("Failed to unmarshal cache metrics data")
```
rutvijmehta-harness, line 50: Don't add any logging for now in case of errors, data collection is best effort. We can consider adding it later for debugging.

ti/savings/savings.go
anurag-harness: Will do it later once we have a need for it

ti/savings/dlc/dlc.go
anurag-harness: Removed

PR #290, opened 2024-06-05
Fix size of log upload blob

rutvijmehta-harness:
Verification in the ticket: https://harness.atlassian.net/browse/CI-12791


vistaarjuneja (approved):
Good catch @rutvijmehta-harness, approved

PR #295, opened 2024-06-18
feat: [CI-12843]: Enable process information in LE logs

raghavharness:
Added following changes.

1. Added process dump in collector if `HARNESS_ENABLE_DEBUG_LOGS` env is set
2. Added timeouts for log upload requests so that backoff can work.

### DIFF
--- a/handler/setup.go
+++ b/handler/setup.go
@@ -22,7 +22,8 @@ import (
 )
 
 var (
-	statsInterval = 30 * time.Second
+	statsInterval          = 30 * time.Second
+	harnessEnableDebugLogs = "HARNESS_ENABLE_DEBUG_LOGS"
 )
 
 // HandleExecuteStep returns an http.HandlerFunc that executes a step
@@ -36,8 +37,11 @@ func HandleSetup(engine *engine.Engine) http.HandlerFunc {
 			WriteBadRequest(w, err)
 			return
 		}
-
-		collector := osstats.New(context.Background(), statsInterval)
+		logProcess := false
+		if val, ok := s.Envs[harnessEnableDebugLogs]; ok && val == "true" {
+			logProcess = true
+		}
+		collector := osstats.New(context.Background(), statsInterval, logProcess)
 
 		setProxyEnvs(s.Envs)
 		state := pipeline.GetState()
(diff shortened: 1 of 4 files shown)


devkimittal:
Can you show one example of the logs?

osstats/collector.go
```diff
@@ -155,6 +165,51 @@ func (s *StatsCollector) get() (*osStat, error) {
 		MemAvailableMB: formatMB(vm.Available), MemUsedMB: formatMB(vm.Used), SwapMemPct: swap.UsedPercent, CPUTotal: s.cpuTotal}, nil
 }
 
+func DumpProcessInfo() error {
+	// Retrieve list of processes
+	processes, err := process.Processes()
```
devkimittal, line 170: given that this process list can be quite huge and since we are running every 10 minutes, might bring down le due to large size of logs or cause slowness. I wonder if we can do some filter or decrease the frequency.
Also init logs might get truncated due to this.

handler/setup.go
```diff
@@ -36,8 +37,11 @@ func HandleSetup(engine *engine.Engine) http.HandlerFunc {
 			WriteBadRequest(w, err)
 			return
 		}
-
-		collector := osstats.New(context.Background(), statsInterval)
+		logProcess := false
+		if val, ok := s.Envs[harnessEnableDebugLogs]; ok && val == "true" {
```
devkimittal, line 41: where is this variable being populated?

osstats/collector.go
raghavharness: Yeah that's why we are printing on demand only. Whenever stage variable is set or no logs are generated for 10 min we are dumping process logs

devkimittal (approved):
Let's check for windows once, just want to make sure the process package exists there as well.

handler/setup.go
raghavharness: didn't get, it will come as stage variable so it'll be part of envs

raghavharness:
checked for windows, mac and linux already. This is go library which is usually supported cross platform.

handler/setup.go
devkimittal: I thought this was coming from runner or ci-manager. clear now.

PR #300, opened 2024-07-17
feat: [CI-13332]:  Add support for dotnet agent

M-Tsur:
## Description

Adding support for .Net to TI
This includes:
* Downloading the agent's archive
* Unzipping the archive
* Configuring the environment variables and writing a configuration file
* Parsing the results - for .Net it can support JUnit and TRX formats, so here we add support for TRX by converting them back to JUnit format in memory and letting the process continue as is

One thing to note - as opposed to other agents, .net is added later to the v2 step, so we don't prevent the use of the v2 step in case the .net agent is failed to download - we would just initiate a warning and won't apply the .net changes.

This change requires the changes from this PR: https://github.com/wings-software/harness-ti/pull/170 but it wouldn't fail if they are missing.

## Testing

* Tested with and without the physical agent being present
* Tested in Devspace and a local dlite setup:
  * Validated the ability to parse the TRX files and convert to JUnit (also tested separately outside of the lite engine)
  * Validated that the agent is being picked up when present. No real testing has been done on validating the actual agent yet as this is on a different task. Currently, the agent is not available, so this code won't do a lot.

## Screenshots
image

image

### DIFF
--- a/pipeline/runtime/runtestsV2.go
+++ b/pipeline/runtime/runtestsV2.go
@@ -39,7 +39,7 @@ const (
 	waitTimeoutInSec  = 30
 	agentV2LinkLength = 3
 	dotNetAgentLinkIndex = 3
-	dotNetAgentProfilerGuid = "{86A1D712-8FAE-4ECD-9333-DB03F62E44FA}"
+	dotNetAgentProfilerGUID = "{86A1D712-8FAE-4ECD-9333-DB03F62E44FA}"
 	dotNetAgentV2Lib    = "net-agent.so"
 	dotNetAgentV2Zip    = "net-agent.zip"
 	dotNetAgentV2Path   = "/dotnet/v2/"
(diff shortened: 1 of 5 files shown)


M-Tsur:
Please feel free to add more reviewers

anurag-harness (commented):
@M-Tsur Can you add a description and testing screenshots?

anurag-harness (commented):
@M-Tsur Please make sure linting also passes

https://harness.drone.io/harness/lite-engine/2406/1/3

anurag-harness (approved):
Can we add UTs?

M-Tsur:
> Can we add UTs?

Sure. I didn't add any as the current tests file for run tests v2 seems a bit lacking in implementations - there are a lot of TODOs there so I wasn't sure of the proper way to really test what I was adding. I will spend some time on it later this week

M-Tsur:
Merging this to make sure we have it in the next release. I will raise a new PR for the unit tests

PR #303, opened 2024-07-24
[feat]: [CI-13579]: Return  Optimization state for Failed Step

anurag-harness:
Currently optimization state is not returned for a failed step. This should be done to make sure cache intel FULL_RUN is saved since cache intel is a FULL_RUN if the restore-cache-harness step fails

Tested on local

FULL_RUN (where restore step fails)

We can see that we are returning the optimization state
Screenshot 2024-07-24 at 12 12 32 PM
Screenshot 2024-07-24 at 12 15 46 PM

Optimized Run
Screenshot 2024-07-24 at 12 14 38 PM
Screenshot 2024-07-24 at 12 16 06 PM

### DIFF
--- a/ti/savings/savings.go
+++ b/ti/savings/savings.go
@@ -65,6 +65,11 @@ func ParseAndUploadSavings(ctx context.Context, workspace string, log *logrus.Lo
 		}
 	}
 
+	// If the Step Fails then we should return the savings state only for restore-cache (CACHE_INTEL). This is to match the current beahavior in K8
+	if !stepSuccess {
+		states = make([]types.IntelligenceExecutionState, 0)
+	}
+
 	// Cache Intel savings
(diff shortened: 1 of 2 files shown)


ti/savings/savings.go
```diff
@@ -65,6 +65,11 @@ func ParseAndUploadSavings(ctx context.Context, workspace string, log *logrus.Lo
 		}
 	}
 
+	// If the Step Fails then we should return the savings state only for restore-cache (CACHE_INTEL). This is to match the current behavior in K8
```
rutvijmehta-harness, line 68: Add a comment stating, no savings parsing should happen after this line of code except cache intel.

rutvijmehta-harness (approved):

ti/savings/savings.go
anurag-harness: Done

PR #305, opened 2024-08-06
feat: [CI-13332]: Enhancements and fixes

M-Tsur:
This PR fixes a small bug in the original implementation, also adds support to alpine which requires a specific agent implementation and also adds support for a new QA environment workflow which allows downloading the agent from a specific qa folder by specifying an environment variable.

### DIFF
--- a/ti/instrumentation/utils.go
+++ b/ti/instrumentation/utils.go
@@ -511,7 +511,13 @@ func DownloadFile(ctx context.Context, path, url string, fs filesystem.FileSyste
 
 func GetV2AgentDownloadLinks(ctx context.Context, config *tiCfg.Cfg, useQAEnv bool) ([]ti.DownloadLink, error) {
 	c := config.GetClient()
-	links, err := c.DownloadLink(ctx, "RunTestV2", runtime.GOOS, runtime.GOARCH, "", "", "qa")
+
+	buildEnv := ""
+	if useQAEnv {
+		buildEnv = "qa"
+	}
+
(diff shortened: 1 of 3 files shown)


anurag-harness (approved):
Please gofmt the changes: https://harness.drone.io/harness/lite-engine/2450/1/3

ti/instrumentation/utils.go
```diff
@@ -509,9 +509,9 @@ func DownloadFile(ctx context.Context, path, url string, fs filesystem.FileSyste
 	return nil
 }
 
-func GetV2AgentDownloadLinks(ctx context.Context, config *tiCfg.Cfg) ([]ti.DownloadLink, error) {
+func GetV2AgentDownloadLinks(ctx context.Context, config *tiCfg.Cfg, useQAEnv bool) ([]ti.DownloadLink, error) {
```
smjt-h, line 512: Why are we useQAEnv

ti/instrumentation/utils.go
```diff
@@ -509,9 +509,9 @@ func DownloadFile(ctx context.Context, path, url string, fs filesystem.FileSyste
...
 	c := config.GetClient()
-	links, err := c.DownloadLink(ctx, "RunTestV2", runtime.GOOS, runtime.GOARCH, "", "", "")
+	links, err := c.DownloadLink(ctx, "RunTestV2", runtime.GOOS, runtime.GOARCH, "", "", "qa")
```
smjt-h, line 514: will we always send "qa" parameter

M-Tsur: You are entirely right. I actually missed pushing that change (it will address the previous comment as well)

PR #306, opened 2024-08-14
Use RunnerResponse http client method

mdmilic:
Added RunnerResponse to StepStatusConfig struct which will be configured by the runner.
Started using new SendRunnerResponse to send Runner response to the new response endpoint.
Tested in localdev by running drone-runner-aws & LE

### DIFF
--- a/pipeline/runtime/step_executor.go
+++ b/pipeline/runtime/step_executor.go
@@ -407,6 +407,7 @@ func (e *StepExecutor) sendStepStatus(r *api.StartStepRequest, response *api.VMT
 
 func (e *StepExecutor) sendStatus(r *api.StartStepRequest, delegateClient *delegate.HTTPClient, response *api.VMTaskExecutionResponse) error {
 	if r.StepStatus.RunnerResponse {
+		logrus.WithField("id", r.ID).Infoln("Sending runner step status")
 		status := client.Success
 		if response.CommandExecutionStatus == api.Failure {
 			status = client.Failure
@@ -430,6 +431,7 @@ func (e *StepExecutor) sendStatus(r *api.StartStepRequest, delegateClient *deleg
 		}
 		return delegateClient.SendRunnerStatus(context.Background(), r.StepStatus.DelegateID, r.StepStatus.TaskID, taskResponse)
 	} else {
+		logrus.WithField("id", r.ID).Infoln("Sending legacy step status")
 		// For legacy backwards compatibility treat timeout as failure
 		if response.CommandExecutionStatus == api.Timeout {
 			response.CommandExecutionStatus = api.Failure
--- a/pipeline/runtime/step_executor.go
+++ b/pipeline/runtime/step_executor.go
@@ -418,6 +418,7 @@ func (e *StepExecutor) sendStatus(r *api.StartStepRequest, delegateClient *deleg
 		jsonData, err := json.Marshal(response)
 		// In case of invalid response data, send failure response
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pipeline/runtime/step_executor.go
```diff
@@ -130,7 +130,7 @@ func (e *StepExecutor) StartStepWithStatusUpdate(ctx context.Context, r *api.Sta
 			e.sendStepStatus(r, &resp)
 			return
 		case <-time.After(defaultStepTimeout):
-			resp = api.VMTaskExecutionResponse{CommandExecutionStatus: api.Failure, ErrorMessage: "step timed out"}
+			resp = api.VMTaskExecutionResponse{CommandExecutionStatus: api.Timeout, ErrorMessage: "step timed out"}
```
vistaarjuneja, line 133: Want to make sure this doesn't cause a regression.
@raghavharness I remember you added this some time back. Do you know if there's a reason we return failure here (any special handling we do on the platform side?)

pipeline/runtime/step_executor.go
```diff
@@ -396,24 +396,58 @@ func setPrevStepExportEnvs(r *api.StartStepRequest) {
 }
 
 func (e *StepExecutor) sendStepStatus(r *api.StartStepRequest, response *api.VMTaskExecutionResponse) {
-	jsonData, err := json.Marshal(response)
-	if err != nil {
-		logrus.WithField("id", r.ID).Errorln("Error marshaling struct:", err)
```
vistaarjuneja, line 401: Let's keep this log line

pipeline/runtime/step_executor.go
vistaarjuneja: ah nevermind, I see you revert it back later

mdmilic: Yes, I wanted timeout only for the new flow.

pipeline/runtime/step_executor.go
```diff
@@ -396,24 +396,58 @@ func setPrevStepExportEnvs(r *api.StartStepRequest) {
...
+		if response.CommandExecutionStatus == api.Failure {
+			status = client.Failure
+		} else if response.CommandExecutionStatus == api.Timeout {
+			status = client.Timeout
+		}
+
+		jsonData, err := json.Marshal(response)
+		// In case of invalid response data, send failure response
+		if err != nil {
+			response.ErrorMessage = "Failed to marshal the response data"
```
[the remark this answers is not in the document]
mdmilic: Actually in general it's not a good idea to expose raw error messages outside your module. That leaks your internals, but also someone might be tempted to start depending on them (and they are not under your control).

pipeline/runtime/step_executor.go
mdmilic: This is still printed here:
`logrus.WithField("id", r.ID).WithError(err).Errorln("failed to send step status")`

new `sendStatus` method (which wraps this marshaling) is now returning err which then gets printed in the above line. If I kept the  `logrus.WithField("id", r.ID).Errorln("Error marshaling struct:", err) `, it would print the error message twice

pipeline/runtime/step_executor.go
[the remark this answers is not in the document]
vistaarjuneja: We should log it somewhere, currently this is not returning this error and it's getting swallowed

pipeline/runtime/step_executor.go
vistaarjuneja: Marshalling errors are getting swallowed by `sendStatus` function

vistaarjuneja (approved):

PR #307, opened 2024-08-16
fix: [CI-13836]: Override registry for internal harness images

raghavharness:


engine/docker/docker.go
```diff
@@ -324,18 +324,37 @@ func (e *Docker) create(ctx context.Context, pipelineConfig *spec.PipelineConfig
 		)
 	}
 
+	originalImage := step.Image
+	overriddenImage := originalImage
+
+	// override image registry for internal images
+	// this is short term solution
+	// override if no auth is present
```
hemanthmantri, line 332: Instead of original, overridden, can we use more descriptive names to indicate `dockerHubImage`, `garImage` or similar?

engine/docker/docker.go
```diff
@@ -324,18 +324,37 @@ func (e *Docker) create(ctx context.Context, pipelineConfig *spec.PipelineConfig
...
+	if isHosted && (step.Auth == nil || step.Auth.Username == "" || step.Auth.Password == "") {
```
hemanthmantri, line 333: For internal images didn't we decide to override regardless of auth being present? Is this auth for the default docker connector (or the connector specified in the step)?

engine/docker/docker.go
```diff
@@ -324,18 +324,37 @@ func (e *Docker) create(ctx context.Context, pipelineConfig *spec.PipelineConfig
...
-		(step.Pull == spec.PullDefault && image.IsLatest(step.Image)) {
-		pullerr := e.pullImageWithRetries(ctx, step.Image, pullopts, output)
+		(step.Pull == spec.PullDefault && image.IsLatest(overriddenImage)) {
+		pullerr := e.pullImageWithRetries(ctx, overriddenImage, pullopts, output)
 		if pullerr != nil {
-			return pullerr
+			// if for some reason overridden image does not work then fallback
+			if overriddenImage != originalImage {
+				selectedImage = originalImage
+				pullerr = e.pullImageWithRetries(ctx, originalImage, pullopts, output)
```
hemanthmantri, line 348: Should you not be using `selectedImage` here? or remove line 347.

engine/docker/image/image.go
```diff
@@ -15,6 +15,13 @@ import (
 	"github.com/docker/distribution/reference"
 )
 
+var (
+	internalImages = []string{"harness/drone-git", "plugins/docker", "plugins/acr", "plugins/ecr", "plugins/gcr",
+		"plugins/gar", "plugins/gcs", "plugins/s3", "harness/sto-plugin", "plugins/artifactory", "plugins/cache",
+		"harness/ssca-plugin", "harness/slsa-plugin", "harness/ssca-compliance-plugin"}
+	garRegistry = "us-docker.pkg.dev/gar-prod-setup/harness-public/"
+)
+
```
hemanthmantri, line 24: Candidate for consts.go or similar.

pipeline/runtime/run.go
```diff
@@ -91,7 +91,10 @@ func executeRunStep(ctx context.Context, f RunFunc, r *api.StartStepRequest, out
 	log := logrus.New()
 	log.Out = out
 
-	exited, err := f(ctx, step, out, r.LogDrone)
+	// stageRuntimeID is only passed for dlite
+	isHosted := r.StageRuntimeID != ""
```
hemanthmantri, line 95: Is this future-proof and reliable way to determine if we are running on Hosted?

engine/docker/docker.go
smjt-h: IMO we should override only for default connector that is used for internal images, for customer overridden connector or customer specified images we need not override the registry

engine/docker/docker.go
smjt-h: selectedImage is used for container creation only because  overriddenImage is re used in case of PullNever or not found cases later

engine/docker/docker.go
raghavharness: Yes if a connector is provided with auth we should keep using it. Only for anonymous auth we need to use GAR

PR #308, opened 2024-08-23
support for overwriting existing files

hingerabhinav:
Currently lite engine skips processing a file if it already exists. We should ideally overwrite the file with the content provided. Runner creates a exec task with some files. 
Updated createFiles func to check if file exists -> update permissions to write it -> log error in chmod -> continue with os.create (it will truncate existing files)

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -19,6 +19,7 @@ import (
 	"github.com/harness/lite-engine/engine/exec"
 	"github.com/harness/lite-engine/engine/spec"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


engine/engine.go
```diff
@@ -181,8 +181,13 @@ func createFiles(paths []*spec.File) error {
 		}
 
 		path := f.Path
+
+		// make the file writable (if it exists)
 		if _, err := os.Stat(path); err == nil {
-			continue
+			if err = os.Chmod(path, 0644); err != nil {
+				return errors.Wrap(err,
```
vistaarjuneja, line 188: let's just log and continue instead of returning

hingerabhinav: done

harshjain12 (approved):

vcalasansh (approved):

PR #310, opened 2024-09-02
[fix]: [CI-14185]: For DLC step only try to upload DLC savings

smjt-h:


anurag-harness (changes requested):
This is not the issue. The problem is not with dlc since we have a unique metadata filename for dlc for each step. We need to make sure that the metadata file used in build cache is parsed only by one step

anurag-harness (approved):

bot2-harness:
This pull request has been automatically closed due to 30 days of inactivity. If you believe this was done in error, please feel free to reopen the PR or reach out to the BT team for further assistance. Thank you for your contributions!

PR #312, opened 2024-09-17
[feat]: [CI-12024]: Gradle Build Cache Savings should only be Parsed Once

anurag-harness:
In this PR once we read a profile report we append it with .processed so that it isnt read again. This ensures that the original file is not processed by multiple steps

Tested on devspace

Without changes we can see that non build cache step after the build cache also is optimized
Screenshot 2024-09-16 at 11 44 24 PM

After changes we can see that the step after the build cache step does not upload any optimization state and we can see that the profile file is not present anymore

### DIFF
--- a/ti/savings/cache/gradle/parser.go
+++ b/ti/savings/cache/gradle/parser.go
@@ -52,12 +52,28 @@ func readHTMLFromFile(filePath string) (*html.Node, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer file.Close()
 
 	content, err := io.ReadAll(file)
 	if err != nil {
+		file.Close() // Close the file in case of error
 		return nil, err
 	}
+
+	// Close the file before renaming it
+	err = file.Close()
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


ti/savings/cache/gradle/parser.go
```diff
@@ -52,12 +52,28 @@ func readHTMLFromFile(filePath string) (*html.Node, error) {
...
+	}
+
+	// Append ".processed" to the file path
+	processedFilePath := filePath + ".processed"
+
+	// Rename the file to append ".processed". This ensures that the original file is not processed multiple times
+	err = os.Rename(filePath, processedFilePath)
+	if err != nil {
+		return nil, err
+	}
```
hemanthmantri, line 75: You're renaming after reading the file. To minimize the race the order should be:

1. Rename the file (and maybe defer os.Remove() too)
2. Read the file contents (using new name)

anurag-harness: Made the changes

anurag-harness: Resolving conversation and merging

PR #319, opened 2025-03-17
feat: [CDE-685]: Destroy stopped plugin containers on suspending a VM

vikyathharekal:
- Delete stopped plugin (identified using labels) containers on suspend
- Create network if not exists (on resuming a VM, we need not create network again)


engine/docker/docker.go
```diff
@@ -565,7 +598,15 @@ func (e *Docker) createNetworkWithRetries(ctx context.Context,
...
-	var err error
+	// Check if the network already exists
+	_, _, err := e.client.NetworkInspectWithRaw(ctx, pipelineConfig.Network.ID, types.NetworkInspectOptions{})
+	if err == nil {
+		return nil
+	} else if !client.IsErrNotFound(err) {
+		// If the error is not "network not found", return it
+		return err
+	}
+
```
raghavharness, line 609: Won't this break CI ? We are not creating network and returning directly?

vikyathharekal: It should not break CI as it will fall into 3rd case above

raghavharness (approved):

